PR #2, opened 2019-09-10
RMSRCBUILD-80: Add hotfix branch support for support branches

Stefan-Ilic:
https://re-motion.atlassian.net/browse/RMSRCBUILD-80


ReleaseProcessScript/tools/PesterTester/release_process_script_flow.Tests.ps1
```diff
@@ -124,7 +128,6 @@ Context "Release-Version Functions called" {
       git checkout -b "develop" --quiet
       Test-Add-Commit
 
-
```
iam-mholle: Unnecessary change

ReleaseProcessScript/tools/Core/main_functions.ps1
```diff
@@ -29,10 +29,10 @@ function Release-Version ()
 
   $CurrentBranchname = Get-Current-Branchname
 
-  if (Is-On-Branch "support/")
+  if (Is-On-Branch "hotfix/")
```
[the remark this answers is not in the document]
Stefan-Ilic: An exception with a useful message is being thrown, I'd say that's fine. There's a test for that in IntegrationTests.Tests.ps1 on line 135

```
    It "throws an exception when attempting to release from support" {
      Initialize-Test "Support-TryRelease"

      { Release-Version } | Should Throw "You have to be on either a 'hotfix/*' or 'release/*' or 'develop' or 'master' branch to release a version."
    }
```

Stefan-Ilic:
For continuity, I'll document this here too:

> This seems to be a separate issue, but I guess it needs to be done first, a hotfix branch must be created manually atm?

Hmm, yes it looks like this needs to be done manually at the moment and the issues states that it needs to be a separate issue, but I don't think it needs to happen before this one. My plan would probably be to merge this one and have all the tests. Then, quickly doing the other issue shouldn't be a problem, since it's just adding some new branches. On a hotfix to support release, you can see that it already creates a new hotfix branch though.

> Looking at the graph in the issue, the first hotfix branch does have a patch version of 0. Is that right?

The graphic attached to the issue was incorrect. I have correct it and on top of that have added the graphic source to the repository (where it belongs imo)

MichaelKetting:
> 
> 
> Not sure about the checked in .pptx . Might be better to use something like https://github.com/mermaid-js/mermaid for flow description and rendering. Tagging @MichaelKetting for this one.

@iam-mholle don't think it's worth the expense right now to evaluate a new flow-chart tool just to not having to check in a pptx file.

ReleaseProcessScript/tools/PesterTester/semver_functions.Tests.ps1
```diff
@@ -27,21 +27,21 @@ Describe "semver_functions" {
...
-      $NextVersions = "1.2.0-rc.5", "1.2.0", "2.0.0-alpha.1", "2.0.0-beta.1", "2.0.0"
+      $NextVersions = "1.2.0", "2.0.0-alpha.1", "2.0.0-beta.1", "2.0.0"
 
       Get-Possible-Next-Versions-Develop $Version | Should Be $NextVersions
     }
 
     It "Get-Possible-Next-Versions-Develop_WithoutPre_ShouldReturnArray" {
       $Version = "1.2.0"
-      $NextVersions = "1.3.0-alpha.1", "1.3.0-beta.1", "1.3.0", "2.0.0-alpha.1", "2.0.0-beta.1", "2.0.0"
+      $NextVersions = "1.3.0-alpha.1", "1.3.0-beta.1", "1.3.0", "2.0.0", "2.0.0-alpha.1", "2.0.0-beta.1"
```
MichaelKetting: please explain new ordering

ReleaseProcessScript/tools/PesterTester/semver_functions.Tests.ps1
```diff
@@ -27,21 +27,21 @@ Describe "semver_functions" {
...
 
       Get-Possible-Next-Versions-Develop $Version | Should Be $NextVersions
     }
 
     It "Get-Possible-Next-Versions-Develop_OnlyMajor_ShouldReturnArray" {
       $Version = "2.0.0"
-      $NextVersions = "2.1.0-alpha.1", "2.1.0-beta.1", "2.1.0", "3.0.0-alpha.1", "3.0.0-beta.1", "3.0.0"
+      $NextVersions = "2.1.0-alpha.1", "2.1.0-beta.1", "2.1.0", "3.0.0", "3.0.0-alpha.1", "3.0.0-beta.1"
```
MichaelKetting: please explain new ordering

ReleaseProcessScript/tools/PesterTester/semver_functions.Tests.ps1
```diff
@@ -27,21 +27,21 @@ Describe "semver_functions" {
 
     It "Get-Possible-Next-Versions-Develop_WithRc_ShouldReturnArray" {
       $Version = "1.2.0-rc.4"
-      $NextVersions = "1.2.0-rc.5", "1.2.0", "2.0.0-alpha.1", "2.0.0-beta.1", "2.0.0"
+      $NextVersions = "1.2.0", "2.0.0-alpha.1", "2.0.0-beta.1", "2.0.0"
```
MichaelKetting: please explain.  why do we even start out with an RC version on develop?

ReleaseProcessScript/tools/PesterTester/semver_functions.Tests.ps1
```diff
@@ -54,42 +54,42 @@ Describe "semver_functions" {
...
+      { Get-Possible-Next-Versions-Develop "1.2.3.alpha-4" } | Should Throw "Your version '1.2.3.alpha-4' does not have a valid format (e.g. 1.2.3-alpha.1)"
+      { Get-Possible-Next-Versions-Develop "1.2.3.rc.4" } | Should Throw "Your version '1.2.3.rc.4' does not have a valid format (e.g. 1.2.3-alpha.1)"
     }
   }
 
-  Context "Get-Possible-Next-Versions-Support" {
-      It "Get-Possible-Next-Versions-Support_WithAlpha_ShouldReturnArray" {
-        $Version = "1.2.0-alpha.4"
-        $NextVersions = "1.2.0-alpha.5", "1.2.0-beta.1", "1.2.0"
+  Context "Get-Possible-Next-Versions-Hotfix" {
```
MichaelKetting: typicallly, the hotfix-branches would not have patch-version = 0 given that patch-version = 0 is normally released from the master and used as the base for the support branch. (unless there's a supportbranch off a supportbranch). would it be helpful for reading the code if we updated the version numbers to reflect this?

ReleaseProcessScript/tools/PesterTester/semver_functions.Tests.ps1
```diff
@@ -54,42 +54,42 @@ Describe "semver_functions" {
...
-      }
+      Get-Possible-Versions-Hotfix $Version $true | Should Be $NextVersions
+    }
 
-      It "Get-Possible-Next-Versions-Support_WithoutPre_ShouldReturnArray" {
-        $Version = "1.2.0"
-        $NextVersions = "1.2.1-alpha.1", "1.2.1-beta.1", "1.2.1"
+    It "Get-Possible-Next-Versions-Hotfix_WithoutPre_ShouldReturnArray" {
+      $Version = "1.2.0"
+      $NextVersions = "1.2.1", "1.2.1-alpha.1", "1.2.1-beta.1"
```
MichaelKetting: please explain new order

ReleaseProcessScript/tools/PesterTester/semver_functions.Tests.ps1
MichaelKetting: I noticed the diagram in https://re-motion.atlassian.net/browse/RMSRCBUILD-80 might have a flaw:

we start support/1.3 off support/1.2.1 and then immediately go to hotfix/1.3.1 instead of hotfix/1.3.0. Of course, strictly speaking, v1.3.0 is not a hotfix, but that could be a detail that's not relevant. we should however ensure that a support branch based off a support branch can start with patch-version = 0.

ReleaseProcessScript/tools/PesterTester/IntegrationTests.Tests.ps1
```diff
@@ -10,319 +10,209 @@
 #There was an Issue that $PSScriptRoot was null in BeforeEach/AfterEach, so we have to cache it here
 $ScriptRoot = $PSScriptRoot
 
-$TestBaseDir = "C:\temp"
+$TestBaseDir = (Get-Item -LiteralPath $env:TEMP).FullName
```
MichaelKetting: we don't have integration tests for the new bits, do we?

ReleaseProcessScript/tools/PesterTester/semver_functions.Tests.ps1
Stefan-Ilic: I can't remember too well, I believe it's just a leftover of experimentation, since the test is green after reverting. I'll just revert this change, then

Stefan-Ilic: I think I know why we changed the ordering. The production code is like this:
```
return $NextPossiblePatch, "$($NextPossiblePatch)-alpha.1", "$($NextPossiblePatch)-beta.1"
```
(It was like that before, we didn't touch that part). I believe we changed the order to reflect this code bit, but I can nonetheless revert the changes, since it doesn't affect the tests

MichaelKetting: I see. I assumed that the order of $NextVersions was included in the assertion. Obviously it is not. In this instance it might be a good idea though? Do we have a place where we assert that the "correct" order of items is printed on the next-version prompt? Aside from that, it's okay to leave the order in the assertion in the updated form.

Also, just in case this is the spot where the list of next-versions is checked. Okay, first we go up and then we offer the 2.0-prerelease versions only as a last resort. this kind of makes sense. So, new list is fine.

Stefan-Ilic:
Closed in favor of https://github.com/re-motion/SharedSource-Build/pull/9

PR #6, opened 2019-12-12
RMSRCBUILD-106: Fix Read-Ancestor-Choice called with a single string instead of an array

iam-mholle:
https://re-motion.atlassian.net/browse/RMSRCBUILD-106

### DIFF
--- a/ReleaseProcessScript/tools/PesterTester/git_base_functions.Tests.ps1
+++ b/ReleaseProcessScript/tools/PesterTester/git_base_functions.Tests.ps1
@@ -195,4 +195,5 @@ Describe "git_base_functions" {
 
       Assert-VerifiableMock
     }
+  }
 }


ReleaseProcessScript/tools/PesterTester/git_base_functions.Tests.ps1
```diff
@@ -175,4 +175,18 @@ Describe "git_base_functions" {
       { Check-Branch-Up-To-Date "master" } | Should Not Throw 
     }
   }
+
+  Context "Get-Ancestor" {
+    It "should print all found ancestors if multiple are found" {
+      git checkout -b "additionalAncestor" --quiet
+      git checkout -b "newBranch" --quiet
+      $ExpectedAncestors = "master", "additionalAncestor"
+      Mock Write-Host -Verifiable -ParameterFilter {$Object -eq "We expected to find one of following Ancestors: 'master additionalAncestor', but found multiple possible Ancestors."}
```
MichaelKetting: Hmm...drop single quote. we're printing multiple individual entries which are only whitespace separated inside a block delimited by a colon and a comma. The extra quotes don't help if there is an actual problem and might imply that the printed text is a single item instead of multiple. The completely correct way would be to place each item into its own quotes but that might be too expensive, codewise in powershell.

ReleaseProcessScript/tools/Core/git_base_functions.ps1
```diff
@@ -267,30 +268,26 @@ function Get-Ancestor ($ExpectedAncestors)
...
     }
   }
 
-  if ($ReturnAncestor -eq $NULL)
+  if ($FoundAncestors.Count -eq 0)
+  {
+    return Read-Host "We expected some of the following Ancestors: '$ExpectedAncestors' but found none. Please enter the name of the Ancestor Branch"
+  }
+
+  if ($FoundAncestors.Count -gt 1)
```
MichaelKetting: switch this and the last statement. returning the single item when -eq 1 would be more structed, going from 0 to 1 to many.

ReleaseProcessScript/tools/Core/git_base_functions.ps1
```diff
@@ -267,30 +268,26 @@ function Get-Ancestor ($ExpectedAncestors)
...
-        $ReturnAncestor = $Ancestor
-      }
+      $FoundAncestors.Add($Ancestor) | Out-Null
     }
   }
 
-  if ($ReturnAncestor -eq $NULL)
+  if ($FoundAncestors.Count -eq 0)
+  {
+    return Read-Host "We expected some of the following Ancestors: '$ExpectedAncestors' but found none. Please enter the name of the Ancestor Branch"
```
MichaelKetting: should this end in a colon? do we always use upper case for words like 'Ancestor' and 'Ancestor Branch' in text output? normally, those would be spelled with lower case.

ReleaseProcessScript/tools/PesterTester/git_base_functions.Tests.ps1
Stefan-Ilic: I managed to do a "correcter" way since I found a StackOverflow answer for it

ReleaseProcessScript/tools/Core/git_base_functions.ps1
Stefan-Ilic: Yes, it looks like words related to "Ancestor" always start with an uppercase letter in our ReleaseProcessScript outputs. I've added the colon, though

ReleaseProcessScript/tools/Core/git_base_functions.ps1
```diff
@@ -267,30 +268,26 @@ function Get-Ancestor ($ExpectedAncestors)
...
-        $ReturnAncestor = $Ancestor
-      }
+      $FoundAncestors.Add($Ancestor) | Out-Null
     }
   }
 
-  if ($ReturnAncestor -eq $NULL)
+  if ($FoundAncestors.Count -eq 0)
+  {
+    return Read-Host "We expected some of the following Ancestors: '$ExpectedAncestors' but found none. Please enter the name of the Ancestor Branch:"
```
MichaelKetting: String-split here, too?

MichaelKetting: Okay, please add a task to cleanup the casing in the script.

PR #8, opened 2020-04-06
RMSRCBUILD-115: Upgrade to NUnit 3

Stefan-Ilic:
### Jira Issue
https://re-motion.atlassian.net/browse/RMSRCBUILD-115

### Breaking Changes

The NuGet package `NUnit.Runners` is not supported anymore. Use `NUnit.ConsoleRunner` instead for getting an NUnit executable. Additionally, running NUnit2 tests requires using the `NUnit.Extension.NUnitV2Driver` NuGet package and setting the MsBuild item `NUnitExtensionsPaths` to the package's root path (e.G. `<NUnitExtensionsPaths Include="$(PackagesDirectory)NUnit.Extension.NUnitV2Driver.3.8.0\"/>`). Running V3 tests while the V2 driver is enabled **is** possible.

Also, all `build.cmd` files need to be upgraded, as Nunit3 requires a newer version of nuget.exe which the hardcoded link inside the `build.cmd` files cannot deliver.

### How to test

It was not possible to provide automated tests, since all changes were MsBuild changes. The existing C# tests are all still green (duh). The easiest way to test this would be to use Remotion-IO and its branch `spike/nunit3` that I created. I personally used it to test everything. Running V2 tests on the re-motion trunk also worked as expected.

### Checklist

Since all changes are MsBuild changes, this checklist is almost completely useless. Nevertheless:

- [x] I have adapted schema files if I made changes to XML Config Files.
- [x] All new files have license headers.
- [x] All new public classes, methods and properties have XML documentation and JetBrains annotations.
- [x] I have identified all breaking changes and marked them appropriately.
- [x] I have followed the style guide to the best of my ability.
- [x] If I added new projects to the solution, I also added them to the corresponding `projects.props` file.

### Remarks

I used this opportunity to upgrade the existing C# tests to NUnit3.

### DIFF
--- a/.nuget/packages.config
+++ b/.nuget/packages.config
@@ -4,6 +4,6 @@
   <package id="DependDB.BuildProcessor.NuGetPreProcessor" version="2.0.0" />
   <package id="MSBuild.Extension.Pack" version="1.5.0" />
   <package id="NuGet.CommandLine" version="3.5.0" />
-  <package id="NUnit.Runners" version="2.6.3" />
+  <package id="NUnit.ConsoleRunner" version="3.11.1" targetFramework="net45" />
   <package id="Remotion.BuildTools.MSBuildTasks" version="1.0.5827.25795" />
 </packages>
--- a/BuildScript.UnitTests/packages.config
+++ b/BuildScript.UnitTests/packages.config
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="NUnit" version="2.6.2" targetFramework="net45" />
+  <package id="NUnit" version="3.12.0" targetFramework="net47" />
   <package id="RhinoMocks" version="3.6.1" targetFramework="net45" />
 </packages>
--- a/BuildScript/BuildTargets/Testing.targets
+++ b/BuildScript/BuildTargets/Testing.targets
@@ -36,25 +36,19 @@
 
   <ItemGroup>
     <NumberOfTestFailuresPerProject Include="0"/>
-    <NumberOfTestErrorsPerProject Include="0"/>
   </ItemGroup>
 
   <!-- Main target; reporting only, calls RunTestsInternal for actual work -->
   <Target Name="RunTests" DependsOnTargets="CreateTestConfigurations;CreateLogDirectory;BuildTestProjects;LogRunTestsStartTime;RunTestsInternal">
     <Error Text="The property 'ConfigurationID' is not set." Condition="'$(ConfigurationID)' == ''" />
     <Error Text="ItemGroup 'NumberOfTestFailuresPerProject' is empty, but should contain test results!" Condition="@(NumberOfTestFailuresPerProject) == ''"/>
-    <Error Text="ItemGroup 'NumberOfTestErrorsPerProject' is empty, but should contain test results!" Condition="@(NumberOfTestErrorsPerProject) == ''"/>
-
     <MSBuild.ExtensionPack.Science.Maths TaskAction="Add" Numbers="@(NumberOfTestFailuresPerProject)">
       <Output PropertyName="_failures" TaskParameter="Result"/>
     </MSBuild.ExtensionPack.Science.Maths>
-    <MSBuild.ExtensionPack.Science.Maths TaskAction="Add" Numbers="@(NumberOfTestErrorsPerProject)">
-      <Output PropertyName="_errors" TaskParameter="Result"/>
-    </MSBuild.ExtensionPack.Science.Maths>
 
     <Error
-      Text="Running Unit-Tests failed because of $(_failures) failures and/or $(_errors) errors in: %0A@(FailedProjects->'%(Identity) (%(Platform)/%(DatabaseSystem)) with %(Failures) failures and %(Errors) errors','%0A')"
-      Condition="$(_failures) &gt; 0 or $(_errors) &gt; 0"/>
+      Text="Running Unit-Tests failed because of $(_failures) failures in: %0A@(FailedProjects->'%(Identity) (%(Platform)/%(DatabaseSystem)) with %(Failures) failures','%0A')"
+      Condition="$(_failures) &gt; 0"/>
 
     <PropertyGroup>
       <_timeTaken>$([System.DateTime]::UtcNow.Subtract($([System.DateTime]::Parse($(RunTestsStartTime)))).TotalMilliseconds.ToString(0))</_timeTaken>
@@ -83,8 +77,6 @@
     <Error Text="The property 'NoDB' is not set." Condition="'$(NoDB)' == ''"/>
     <Error Text="The property 'NoBrowser' is not set." Condition="'$(NoBrowser)' == ''"/>
     <Error Text="The property 'NUnitToolPath' is not set." Condition="'$(NUnitToolPath)' == ''"/>
-    <Error Text="The property 'NunitFolderPathWithDefaultConfiguration' is not set." Condition="'$(NunitFolderPathWithDefaultConfiguration)' == ''"/>
-    <Error Text="The property 'NunitFolderPathWithoutDotNet35Dependency' is not set." Condition="'$(NunitFolderPathWithoutDotNet35Dependency)' == ''"/>
 
     <MSBuild.ExtensionPack.Framework.MsBuildHelper TaskAction="GetItemCount" InputItems1="@(_testOutputFiles)">
       <Output TaskParameter="ItemCount" PropertyName="_testOutputFilesCount"/>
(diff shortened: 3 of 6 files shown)


BuildScript.UnitTests/packages.config
```diff
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="NUnit" version="2.6.2" targetFramework="net45" />
+  <package id="NUnit" version="3.12.0" targetFramework="net47" />
```
MichaelKetting, line 3: This means we can only use this once all build agents are updated to .net 4.7 -> dependency on the docker-task @iam-mholle

.nuget/packages.config
```diff
@@ -4,6 +4,7 @@
   <package id="DependDB.BuildProcessor.NuGetPreProcessor" version="2.0.0" />
   <package id="MSBuild.Extension.Pack" version="1.5.0" />
   <package id="NuGet.CommandLine" version="3.5.0" />
-  <package id="NUnit.Runners" version="2.6.3" />
+  <package id="NUnit.ConsoleRunner" version="3.11.1" targetFramework="net45" />
```
MichaelKetting, line 7: > Also, all build.cmd files need to be upgraded, as Nunit3 requires a newer version of nuget.exe which the hardcoded link inside the build.cmd files cannot deliver.

Okay, switching to https://dist.nuget.org/win-x86-commandline/latest/nuget.exe is probably not an issue in build.cmd. Also, which version exactly is the minimum version? We will need to ensure that TeamCity also provides the correct minimum version when executing the build. Currently, we're using NuGet 4.5.1

BuildScript/Main/Remotion.build
```diff
@@ -25,13 +25,17 @@
...
-    <NUnitToolPath Condition="'$(NUnitToolPath)' == ''">$(PackagesDirectory)NUnit.Runners.2.6.3\tools\</NUnitToolPath>
+    <NUnitToolPath Condition="'$(NUnitToolPath)' == ''">$(PackagesDirectory)NUnit.ConsoleRunner.3.11.1\tools\</NUnitToolPath>
     <NuGetToolPath Condition="'$(NuGetToolPath)' == ''">$(PackagesDirectory)NuGet.CommandLine.3.5.0\tools\</NuGetToolPath>
     <NuGetForMSBuildPath Condition="'$(NuGetForMSBuildPath)' == ''">$(PackagesDirectory)NuGet.for.MSBuild.1.4.3\build\</NuGetForMSBuildPath>
     <RemotionBuildTasksPath Condition="'$(RemotionBuildTasksPath)' == ''">$(PackagesDirectory)Remotion.BuildTools.MSBuildTasks.1.0.6862.28795\tools\</RemotionBuildTasksPath>
     <RemotionBuildScriptPath Condition="'$(RemotionBuildScriptPath)' == ''">$(PackagesDirectory)Remotion.BuildScript.2.0.0-alpha020\BuildTargets\</RemotionBuildScriptPath>
   </PropertyGroup>
 
+  <ItemGroup>
+    <NUnitExtensionsPaths Include="$(PackagesDirectory)NUnit.Extension.NUnitV2Driver.3.8.0\"/>
```
MichaelKetting, line 36: We want to do an opt-in to the backward-compatible build, correct. Or do we want to, for now, deliver the Nunit2-compat to all projects regardless of their nunit2 requirements?

BuildScript.UnitTests/packages.config
MichaelKetting: We have 3.12  here and 3.11 when it comes to the NUnit runner. Is that an oversight?

BuildScript/BuildTargets/Testing.targets
```diff
@@ -36,25 +36,19 @@
...
   <ItemGroup>
     <NumberOfTestFailuresPerProject Include="0"/>
-    <NumberOfTestErrorsPerProject Include="0"/>
   </ItemGroup>
 
   <!-- Main target; reporting only, calls RunTestsInternal for actual work -->
   <Target Name="RunTests" DependsOnTargets="CreateTestConfigurations;CreateLogDirectory;BuildTestProjects;LogRunTestsStartTime;RunTestsInternal">
     <Error Text="The property 'ConfigurationID' is not set." Condition="'$(ConfigurationID)' == ''" />
     <Error Text="ItemGroup 'NumberOfTestFailuresPerProject' is empty, but should contain test results!" Condition="@(NumberOfTestFailuresPerProject) == ''"/>
-    <Error Text="ItemGroup 'NumberOfTestErrorsPerProject' is empty, but should contain test results!" Condition="@(NumberOfTestErrorsPerProject) == ''"/>
```
MichaelKetting, line 46: is the errors-concept gone in NUnit 3? Please link docs for review/know-how building

BuildScript/BuildTargets/Testing.targets
```diff
@@ -36,25 +36,19 @@
...
       <Output PropertyName="_failures" TaskParameter="Result"/>
     </MSBuild.ExtensionPack.Science.Maths>
-    <MSBuild.ExtensionPack.Science.Maths TaskAction="Add" Numbers="@(NumberOfTestErrorsPerProject)">
-      <Output PropertyName="_errors" TaskParameter="Result"/>
-    </MSBuild.ExtensionPack.Science.Maths>
 
     <Error
-      Text="Running Unit-Tests failed because of $(_failures) failures and/or $(_errors) errors in: %0A@(FailedProjects->'%(Identity) (%(Platform)/%(DatabaseSystem)) with %(Failures) failures and %(Errors) errors','%0A')"
-      Condition="$(_failures) &gt; 0 or $(_errors) &gt; 0"/>
+      Text="Running Unit-Tests failed because of $(_failures) failures in: %0A@(FailedProjects->'%(Identity) (%(Platform)/%(DatabaseSystem))','%0A')"
```
MichaelKetting, line 50: why drop the %Failures?

BuildScript/BuildTargets/Testing.targets
```diff
@@ -127,16 +119,31 @@
...
       <_mergedTestCategoriesToExclude>$(_mergedTestCategoriesToExclude.Trim(','))</_mergedTestCategoriesToExclude>
 
-      <_excludeCategoriesCallArgumentString></_excludeCategoriesCallArgumentString>
-      <_excludeCategoriesCallArgumentString Condition="'$(_mergedTestCategoriesToExclude)' != ''">/exclude=$(_mergedTestCategoriesToExclude)</_excludeCategoriesCallArgumentString>
+      <_mergedTestCategoriesToInclude>$(TestCategoriesToInclude.Replace(';', ',')),$(_includeCategoriesInTestConfiguration)</_mergedTestCategoriesToInclude>
+      <_mergedTestCategoriesToInclude>$(_mergedTestCategoriesToInclude.Trim(','))</_mergedTestCategoriesToInclude>
     </PropertyGroup>
 
+    <ItemGroup>
+      <SplitCategoriesToExclude Include="$(_mergedTestCategoriesToExclude.Split(','))"/>
```
MichaelKetting, line 127: is those private state or shared between targets. if private, please use _lowerCase syntax. Also, please clear the itemgroups before usage (see other code-places)

BuildScript/BuildTargets/Testing.targets
```diff
@@ -127,16 +119,31 @@
...
+    </PropertyGroup>
 
-      <_includeCategoriesCallArgumentString></_includeCategoriesCallArgumentString>
-      <_includeCategoriesCallArgumentString Condition="'$(_mergedTestCategoriesToInclude)' != ''">/include=$(_mergedTestCategoriesToInclude)</_includeCategoriesCallArgumentString>
+    <PropertyGroup>
+      <_nunitTestFilter></_nunitTestFilter>
+      <_nunitTestFilter Condition="$(_mergedTestCategoriesToExclude)!='' And $(_mergedTestCategoriesToInclude)!=''">$(_excludeCategoriesCallArgumentString) &amp;&amp; $(_includeCategoriesCallArgumentString)</_nunitTestFilter>
+      <_nunitTestFilter Condition="$(_mergedTestCategoriesToExclude)!='' And $(_mergedTestCategoriesToInclude)==''">$(_excludeCategoriesCallArgumentString)</_nunitTestFilter>
+      <_nunitTestFilter Condition="$(_mergedTestCategoriesToExclude)=='' And $(_mergedTestCategoriesToInclude)!=''">$(_includeCategoriesCallArgumentString)</_nunitTestFilter>
+      <_nunitTestFilter Condition="$(_nunitTestFilter)!=''">--where "$(_nunitTestFilter)"</_nunitTestFilter>
```
MichaelKetting, line 146: should we add a comment sample that shows the final output for include+exclude with multiple categorires? It's a bit mroe involved than before.

BuildScript/BuildTargets/Testing.targets
```diff
@@ -201,80 +212,55 @@ $(_dockerImageName) $(_nunitCallArgumentString)
...
-    <ItemGroup>
-      <_nunitDestinationFolderPath Include="$(NunitFolderPathWithDefaultConfiguration)">
-        <SupportedRuntimeConfiguration></SupportedRuntimeConfiguration>
-      </_nunitDestinationFolderPath>
-
-      <_nunitDestinationFolderPath Include="$(NunitFolderPathWithoutDotNet35Dependency)">
-        <SupportedRuntimeConfiguration>&lt;supportedRuntime version=&quot;v4.0.30319&quot; /&gt;</SupportedRuntimeConfiguration>
-      </_nunitDestinationFolderPath>
-    </ItemGroup>
+    <WriteLinesToFile
```
MichaelKetting, line 252: What's that one doing? it's just creating a file with the contents
> *.dll

?

BuildScript.UnitTests/packages.config
Stefan-Ilic: No, it's just the way NUnit released their stuff. The newest NUnit "types" are at 3.12 while the newest NUnit console runner is at 3.11.

.nuget/packages.config
Stefan-Ilic: The error message was:
```
The 'NUnit 3.12.0' package requires NuGet client version '2.12' or above, but the current NuGet version is '2.8.60717.93'.
```

Meaning we should be fine

BuildScript/Main/Remotion.build
Stefan-Ilic: I remember you saying we can always ship the NUnit2 compatibility if having it enabled doesn't affect running V3 tests, and it doesn't so I added it there. I could remove it though, I don't think it matters much. I also decided against always shipping [NUnit.Extension.TeamCityEventListener](https://www.nuget.org/packages/NUnit.Extension.TeamCityEventListener/) since it wasn't "essential" to the builld, even though we'd probably want it in every project

BuildScript/BuildTargets/Testing.targets
Stefan-Ilic: Oversight, thanks for noticing

BuildScript/Main/Remotion.build
MichaelKetting: that nunit-compat question was in regard to "can we do this, if we have to". But if we don't need to always include it, it's one moving part less we have to worry about.

The teamcity-one: will need to have more information to make a good descison on include or not

.nuget/packages.config
MichaelKetting: yes.

BuildScript/BuildTargets/Testing.targets
Stefan-Ilic: `nunit.engine.dll` in `NUnitToolPath` looks for `.addins` file next to it (meaning in the same directory) to include NUnit extensions (like the V2 driver) when running tests. What I do here is create a custom `.addins` file that includes all `.dll` files inside the same directory, place it inside `NUnitToolPath` then I copy `NUnitToolPath` and `NUnitExtensionsPaths` to the same temporary directory, meaning NUnit itself and all NUnit extensions are together in one place.

NUnit then scans all `.dll` files if they are extensions and includes them in the test run if they are.

BuildScript/BuildTargets/Testing.targets
Stefan-Ilic: I just noticed that the `errors` attribute was gone from the result xml. Assertion failures and unexpected exceptions (in tests and setup/teardown methods) now seem to all be grouped as failures (according to the things I tried).

[This](https://github.com/nunit/docs/wiki/XML-Formats) and [this](https://github.com/nunit/docs/wiki/Test-Result-XML-Format) explain the new format. It says that an `<assertion>` tag has an attribute `result` that may be  Inconclusive, Passed, Warning, Failed or Error, but it looks like we can't just easily get the number of errors and failures (which is probably fine since all we did was group them anyway)

MichaelKetting: Okay, good to know. Please try to trigger somehting that resulted in an error in nunit2 (if reasonably possible) and then try this again in nunit3 and nunit3 with nunit2-compat to see if the failure count covers everything or if we need to improve the error detection.

Stefan-Ilic: Tested in IO: 3 errors and 2 failures in V2 with the old runner were always 5 failures with the new runner, with v2 compatibility and without.

Looks like we're fine

Stefan-Ilic:
> see comment regarding testing. if that doesn't create new issues, we're done?

I need to run big remotion's nunit3 branch on TC with these BuildScript changes. If that works (and I'd expect it to work) we're done

MichaelKetting:
> 
> 
> > see comment regarding testing. if that doesn't create new issues, we're done?
> 
> I need to run big remotion's nunit3 branch on TC with these BuildScript changes. If that works (and I'd expect it to work) we're done

This refers to exercising failing tests and errors in addition to green tests.

PR #9, opened 2020-04-24
RMSRCBUILD-80: Add hotfix branch support for support branches

Stefan-Ilic:
re-motion.atlassian.net/browse/RMSRCBUILD-80

This PR was re-opened with a new branch that now only contains relevant hotfix changes.

PR #10, opened 2020-04-24
RMSRCBUILD-117: Fix incorrect ReleaseProcessScript Tests

Stefan-Ilic:
https://re-motion.atlassian.net/browse/RMSRCBUILD-117

This PR builds on https://github.com/re-motion/SharedSource-Build/pull/9. You should probably review that one first.

### DIFF
--- a/ReleaseProcessScript/packages.config
+++ b/ReleaseProcessScript/packages.config
@@ -1,4 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
+  <package id="GitForWindows" version="2.26.2" targetFramework="net461" />
   <package id="Remotion.BuildTools.MSBuildTasks" version="1.0.6862.28795" />
 </packages>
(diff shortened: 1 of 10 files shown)


ReleaseProcessScript/packages.config
```diff
@@ -1,4 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
+  <package id="GitForWindows" version="2.26.2" targetFramework="net461" />
```
MichaelKetting, line 3: What's up with this dependency? Up until now, we did not need a pre-defined git-dependency but used the one from the client project's context.

ReleaseProcessScript/tools/Core/main_helper_functions.ps1
```diff
@@ -2,12 +2,12 @@ function Parse-Version-From-ReleaseBranch ($Branchname)
 {
   $SplitBranchname = $Branchname.Split("/v")
 
-  if ($SplitBranchname.Length -ne 3)
```
MichaelKetting, line 5: oh, look, there's 3 off-by-one errors in this file :)

ReleaseProcessScript/packages.config
MichaelKetting: Okay, from what I can tell, the tests are part of the production code. the NuGet dependency remain a problem specified this way because it will be included in the finished package. there's a developmentonly-flag available that should take care of business.

PR #12, opened 2020-04-24
RMSRCBUILD-119: Upgrade to BuildScript 2.0.0-alpha022

Stefan-Ilic:
https://re-motion.atlassian.net/browse/RMSRCBUILD-119

This PR builds on https://github.com/re-motion/SharedSource-Build/pull/11. You should probably review that one first.


Build/Customizations/Projects.props
```diff
@@ -1,7 +1,26 @@
...
-	<ItemGroup>
+  <PropertyGroup>
+    <NormalTestConfiguration Condition="'$(NormalTestConfiguration)' == ''">
+      Win_NET48  + NET48  + NoBrowser + NoDB + Debug   + x86;
+      Win_NET48  + NET48  + NoBrowser + NoDB + Release + x86;
+      Win_NET48  + NET48  + NoBrowser + NoDB + Debug   + x64;
+      Win_NET48  + NET48  + NoBrowser + NoDB + Release + x64;
+
+      <!-- Local-->
+      LocalMachine  + NET48  + NoBrowser + SqlServerDefault + Debug + x86;
```
iam-mholle, line 11: Shouldn't this be `NoDB`?

Build/Customizations/Projects.props
```diff
@@ -1,7 +1,26 @@
...
+
+      <!-- Compatibility Mode -->
+      Win_NET48  + NET472 + NoBrowser + NoDB + Release + x64;
+      Win_NET472 + NET472 + NoBrowser + NoDB + Release + x64;
+    </NormalTestConfiguration>
```
iam-mholle, line 16: Isn't this more than needed?

Build/Remotion.build
```diff
@@ -29,9 +29,18 @@
...
   </PropertyGroup>
 
+  <ItemGroup>
+    <SupportedBrowsers Include="Chrome;InternetExplorer;Firefox;Edge"/>
+    <SupportedPlatforms Include="x86;x64"/>
+    <SupportedDatabaseSystems Include="SqlServerDefault;SqlServer2012;SqlServer2014;SqlServer2016;SqlServer2017;SqlServer2019"/>
+    <SupportedExecutionRuntimes Include="Win_NET472=$(DockerImageDotNetFrameworkRuntime4_7);Win_NET48=$(DockerImageDotNetFrameworkRuntime4_8)"/>
+    <SupportedTargetRuntimes Include="NET48;NET472;NET462;"/><!-- Remove NET462 once NET48 is installed on build agents -->
+    <SupportedConfigurationIDs Include="Debug;Release"/>
+  </ItemGroup>
```
iam-mholle, line 42: Must this always be specified? even without web and db tests?

Build/Customizations/Projects.props
MichaelKetting: The comment should be updated, aside from that, no that's what we want. the last two configs + win_Net48+NET48+Release+x64 represent the production scenarios for the deployed binaries and platform. The x86 and Debug configs represent 100% coverage of the code paths in the binaries.

Build/Remotion.build
Stefan-Ilic: I suppose it's clean if we still specify `SupportedBrowsers`and `SupportedDatabaseSystems`, but they should probably be empty if we support none

Build/Remotion.Server.build
```diff
@@ -107,4 +107,10 @@
 
   <Import Project="$(CustomizationsDirectory)Server\*.targets" />
 
+  <Target Name="CheckBuildLocationProperties" >
+  </Target>
+
+  <Target Name="CheckBuildConfigurationProperties" >
+  </Target>
+
```
iam-mholle, line 115: Why is this needed?

Stefan-Ilic: The target `CheckProperties` defined in the same file requires those targes. I was unsure if I should add these empty targets and leave them as an "extension point" or remove the dependency on them altogether

iam-mholle: I see. Is this the right issue to make this change? Is this generally even needed? I guess that's one for @MichaelKetting

Stefan-Ilic: It literally wouldn't compile without them, so we either include these empty targets or remove the dependencies on them

Build/Remotion.Server.build
```diff
@@ -89,7 +89,7 @@
     <Error Text="The property 'BuildNumber' is not set." Condition="'$(BuildNumber)' == ''" />
 
     <PropertyGroup>
-      <Version>$(Version)-x.1.$(BuildNumber)</Version>
+      <Version>$(Version).$(BuildNumber)</Version> <!-- I had to change it since the old way threw an error -->
```
Stefan-Ilic, line 92: The BuildScript has a check if the version is in a correct SemVer format as per http://semver.org/spec/v2.0.0.html, and adding the `-x.1.$(BuildNumber)` seems to break the pattern as "2.0.0-alpha.22-x.1.1337" is not correct SemVer.

MichaelKetting: Yes, there's a way to deal with this. the version.props must not have "-alpha.20" in the entry.
https://github.com/re-motion/SharedSource-Build/blob/develop/Build/Customizations/Version.props#L12

This is also an issue when doing a no-prerelease patch of the version.props file. The current logic adds the prerlease info to the version number. => please revert.

Build/Remotion.Server.build
MichaelKetting: I don't understand. this worked before, the targets are defined in Properties.targets and loaded via Remotion.build import-startement. This should not be needed.

Build/Customizations/Projects.props
```diff
@@ -1,7 +1,26 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-	<ItemGroup>
+  <PropertyGroup>
+    <NormalTestConfiguration Condition="'$(NormalTestConfiguration)' == ''">
+      Win_NET48  + NET48  + NoBrowser + NoDB + Debug   + x86;
```
MichaelKetting, line 5: Hmm.. in the build script's own tests we can actually remove all but one entry since it's powershell we're running anyway. The powershell environment is always using NET45 as target framework, regardless of the platform.

Build/Customizations/UpdateBuildScriptValue.targets
```diff
@@ -19,7 +19,6 @@
   <PropertyGroup>
     <RemotionBuildFile Condition="'$(RemotionBuildFile)' == ''">$(SolutionDirectory)\BuildScript\Main\Remotion.build</RemotionBuildFile>
     <AssemblyNuGetVersion></AssemblyNuGetVersion>
-    <Version></Version>
```
MichaelKetting, line 22: change why?

Build/Remotion.Server.build
```diff
@@ -89,7 +89,7 @@
     <Error Text="The property 'BuildNumber' is not set." Condition="'$(BuildNumber)' == ''" />
 
     <PropertyGroup>
-      <Version>$(Version)-x.1.$(BuildNumber)</Version>
+      <Version>$(Version).$(BuildNumber)</Version>
```
MichaelKetting, line 92: revert

Build/Remotion.Server.build
Stefan-Ilic: You're right, I reverted the change and it still works. I must have done something else wrong before

Build/Customizations/Projects.props
Stefan-Ilic: So the only test configuration besides the local one would be `Win_NET45  + NET45  + NoBrowser + NoDB + Release + x64;`?

MichaelKetting: WIN_NET48 + NET45 or WIN_NET47 + NET45

There's not NET45 docker images and the systems are all updated to newer .NET versions.

Build/Customizations/UpdateBuildScriptValue.targets
Stefan-Ilic: To me it looks like this: *.targets files are imported after *.props files, so this overrides the used version to Empty, making the build fail. I'm guessing this is a bug that nobody noticed?

Stefan-Ilic: I'm unsure of what setting Version to empty was supposed to achieve in the first place

MichaelKetting: > 
> 
> I'm unsure of what setting Version to empty was supposed to achieve in the first place

Just declaring the "variable" / "parameter". It wasn't meant to set to blank.

MichaelKetting: > 
> 
> To me it looks like this: *.targets files are imported after *.props files (Version is set to 2.0.0 in Version.props), so this overrides the used version to Empty, making the build fail. I'm guessing this is a bug that nobody noticed?

Yes, props before code, then targets at the end is the import order normally implemented by MSBuild scripts.

Just noticed, we're talking about something from the build script customizations in the build for this repository. yeah, okay, probably something weird/incorrect going on. Just make it good :)

PR #13, opened 2020-04-28
RMSRCBUILD-120: NUnit3 [OneTimeTeardown] errors are not being reported as test failures

Stefan-Ilic:
Questions you may have:

**Are the tests that ran in a fixture with a failed [OneTimeTeardown] marked as failed?**
No, Teamcity does not show them as failed. NUnit does, however, print an error message with the csproj file containing the failed method, so identifying it is not hard. With this change, we now print out an extra error message and fail the build if such a method fails, and NUnit's error output is very close to our error message so it will be seen. Rider (and probably Resharper) DO mark the entire fixture as failed, so inside the IDE this was never a problem.

**What happens if we have "normal" assert failures and failing OneTimeTeardowns in a single build?**
The build will fail as expected, but the tests with the failing OneTimeTeardown won't be marked as failed either. The above mentioned NUnit error message WILL be present in the build log, but it might not be prominent. If these assert failures are fixed and another build is started, our new, prominent error message will be printed.

**Do [SetUp], [OneTimeSetUp] and [Teardown] cause the same problems?**
No, it seems that a test is only _not_ marked as failed if OneTimeTeardown failed. If any of the other methods fail, the tests were always marked as failed in my testings.

### DIFF
--- a/BuildScript/BuildTargets/Testing.targets
+++ b/BuildScript/BuildTargets/Testing.targets
@@ -48,7 +48,11 @@
 
     <Error
       Text="Running Unit-Tests failed because of $(_failures) failures in: %0A@(FailedProjects->'%(Identity) (%(Platform)/%(DatabaseSystem)) with %(Failures) failures','%0A')"
-      Condition="$(_failures) &gt; 0"/>
+      Condition="$(_failures) &gt; 0 and '$(_nunitIsSuccess)' == 'False'"/>
+
+    <Error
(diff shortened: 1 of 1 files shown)


MichaelKetting:
Those FAQs should be in the NUnit3-Migrator project's upgrade documentation.

BuildScript/BuildTargets/Testing.targets
```diff
@@ -48,7 +48,11 @@
 
     <Error
       Text="Running Unit-Tests failed because of $(_failures) failures in: %0A@(FailedProjects->'%(Identity) (%(Platform)/%(DatabaseSystem)) with %(Failures) failures','%0A')"
-      Condition="$(_failures) &gt; 0"/>
+      Condition="$(_failures) &gt; 0 and '$(_nunitIsSuccess)' == 'False'"/>
```
MichaelKetting, line 51: Do we need this second check here? if there's failures, we don't care about nunitisuccess.

PR #14, opened 2020-05-08
RMSRCBUILD-123: Running NUnit2 tests with the NUnit3 console runner produces incorrect assembly name in result XML

Stefan-Ilic:
https://re-motion.atlassian.net/browse/RMSRCBUILD-123

### DIFF
--- a/BuildScript/BuildTargets/Testing.targets
+++ b/BuildScript/BuildTargets/Testing.targets
@@ -242,6 +242,16 @@ $(_dockerImageName) $(_nunitCallArgumentString)
       <Output TaskParameter="Value" PropertyName="_nunitResult" />
     </MSBuild.ExtensionPack.Xml.XmlFile>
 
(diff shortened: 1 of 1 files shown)


BuildScript/BuildTargets/Testing.targets
```diff
@@ -242,6 +242,16 @@ $(_dockerImageName) $(_nunitCallArgumentString)
       <Output TaskParameter="Value" PropertyName="_nunitResult" />
     </MSBuild.ExtensionPack.Xml.XmlFile>
 
+    <!--  NUnit2 tests display the assembly path in both the name and fullname attribute  -->
```
MichaelKetting, line 245: Ah yes, nice workaround! I was thinking we'd have to extract the value, transform it, and then write it back. But given we already have the value we are looking for available locally, we can simplify this this :)

PR #15, opened 2020-05-12
RMSRCBUILD-122: Testing.targets throws a NullReferenceException when attempting to test an assembly without tests

Stefan-Ilic:
https://re-motion.atlassian.net/browse/RMSRCBUILD-122

### DIFF
--- a/BuildScript/BuildTargets/Testing.targets
+++ b/BuildScript/BuildTargets/Testing.targets
@@ -228,6 +228,10 @@ $(_dockerImageName) $(_nunitCallArgumentString)
       <Output TaskParameter="Value" PropertyName="_testTime" />
     </MSBuild.ExtensionPack.Xml.XmlFile>
 
+    <PropertyGroup>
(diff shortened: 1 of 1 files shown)


BuildScript/BuildTargets/Testing.targets
```diff
@@ -228,6 +228,10 @@ $(_dockerImageName) $(_nunitCallArgumentString)
       <Output TaskParameter="Value" PropertyName="_testTime" />
     </MSBuild.ExtensionPack.Xml.XmlFile>
 
+    <PropertyGroup>
+      <_testTime Condition="'$(_testTime)' == ''">0</_testTime>
+    </PropertyGroup>
+
```
iam-mholle, line 234: Isn't this a separate issue?

Stefan-Ilic: Funnily enough, no. This was the reason why the `NullReferenceException` occurred, the duration attribute was not defined in the result XML when no tests were found.

The other extra handling is to circumvent the fact that the XML is also marked as "Failed" if no tests are found.

PR #17, opened 2020-05-18
RMSRCBUILD-125: Expose DockerImage information with EnforcedLocalMachine Builds

Stefan-Ilic:
https://re-motion.atlassian.net/browse/RMSRCBUILD-125

These changes are meant to be entirely additive, test configurations can still enforce running on local machines without exposing a docker image. The Remotion trunk web test configurations would look like this:

```
Chrome           + NET462 + Debug   + x86 + NoDB + Win_NET48  + EnforcedLocalMachine;
Firefox          + NET462 + Release + x64 + NoDB + Win_NET472 + EnforcedLocalMachine
InternetExplorer + NET462 + Release + x64 + NoDB + Win_NET48  + EnforcedLocalMachine;
Edge             + NET462 + Release + x64 + NoDB + Win_NET48  + EnforcedLocalMachine;
```

### DIFF
--- a/BuildScript/BuildTasks/ExecutionRuntime.cs
+++ b/BuildScript/BuildTasks/ExecutionRuntime.cs
@@ -25,7 +25,6 @@ namespace Remotion.BuildScript.BuildTasks
     public string Value { get; }
     public string DockerImage { get; }
 
-
     public ExecutionRuntime (string key, string value, bool useDocker, string dockerImage)
     {
       Key = key;
--- a/BuildScript/BuildTasks/ExecutionRuntime.cs
+++ b/BuildScript/BuildTasks/ExecutionRuntime.cs
@@ -23,12 +23,15 @@ namespace Remotion.BuildScript.BuildTasks
     public bool UseDocker { get; }
     public string Key { get; }
     public string Value { get; }
+    public string DockerImage { get; }
 
-    public ExecutionRuntime (string key, string value, bool useDocker)
+
+    public ExecutionRuntime (string key, string value, bool useDocker, string dockerImage)
     {
       Key = key;
       Value = value;
       UseDocker = useDocker;
+      DockerImage = dockerImage;
     }
   }
 }
--- a/BuildScript/BuildTasks/ExecutionRuntime.cs
+++ b/BuildScript/BuildTasks/ExecutionRuntime.cs
@@ -32,5 +32,13 @@ namespace Remotion.BuildScript.BuildTasks
       UseDocker = useDocker;
       DockerImage = dockerImage;
     }
+
+    public ExecutionRuntime (string sameKeyAndValue, bool useDocker, string dockerImage)
+    {
+      Key = sameKeyAndValue;
+      Value = sameKeyAndValue;
+      UseDocker = useDocker;
+      DockerImage = dockerImage;
+    }
   }
 }
--- a/BuildScript/BuildTasks/ExecutionRuntime.cs
+++ b/BuildScript/BuildTasks/ExecutionRuntime.cs
@@ -32,13 +32,5 @@ namespace Remotion.BuildScript.BuildTasks
       UseDocker = useDocker;
       DockerImage = dockerImage;
     }
-
-    public ExecutionRuntime (string sameKeyAndValue, bool useDocker, string dockerImage)
-    {
-      Key = sameKeyAndValue;
-      Value = sameKeyAndValue;
-      UseDocker = useDocker;
-      DockerImage = dockerImage;
-    }
   }
 }
--- a/BuildScript/BuildTasks/MetadataValueDictionary.cs
+++ b/BuildScript/BuildTasks/MetadataValueDictionary.cs
@@ -74,15 +74,8 @@ namespace Remotion.BuildScript.BuildTasks
 
       var intersection = _rawDictionary.Keys.Intersect (enumeratedPossibleKeys, _dictionaryStringComparer).ToArray();
 
-      if (intersection.Length == 0)
-      {
-        throw createEmptySequenceException();
-      }
-      else
-      {
-        foreach (var key in intersection)
-          yield return new KeyValuePair<string, string> (key, _rawDictionary[key]);
-      }
+      foreach (var key in intersection)
+        yield return new KeyValuePair<string, string> (key, _rawDictionary[key]);
     }
   }
 }
--- a/BuildScript/BuildTasks/MetadataValueDictionary.cs
+++ b/BuildScript/BuildTasks/MetadataValueDictionary.cs
@@ -72,10 +72,9 @@ namespace Remotion.BuildScript.BuildTasks
     {
       var enumeratedPossibleKeys = possibleKeys.ToArray();
 
-      var intersection = _rawDictionary.Keys.Intersect (enumeratedPossibleKeys, _dictionaryStringComparer).ToArray();
-
-      foreach (var key in intersection)
-        yield return new KeyValuePair<string, string> (key, _rawDictionary[key]);
+      return _rawDictionary.Keys
+          .Intersect (enumeratedPossibleKeys, _dictionaryStringComparer)
+          .Select (key => new KeyValuePair<string, string> (key, _rawDictionary[key]));
     }
   }
 }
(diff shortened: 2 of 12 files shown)


BuildScript/BuildTasks/ExecutionRuntime.cs
```diff
@@ -23,12 +23,15 @@ public class ExecutionRuntime
     public bool UseDocker { get; }
     public string Key { get; }
     public string Value { get; }
+    public string DockerImage { get; }
 
-    public ExecutionRuntime (string key, string value, bool useDocker)
+
```
iam-mholle, line 28:
```suggestion
```

BuildScript/BuildTasks/TestConfigurationsFactory.cs
```diff
@@ -144,11 +144,46 @@ private string GetPlatform (IEnumerable<string> configurationItems)
...
+
+      var kvp = _supportedExecutionRuntimes.Single (enumeratedConfigurationItems, () => CreateMissingConfigurationItemException ("execution runtime"));
 
       var shouldUseDocker = ShouldUseDocker (kvp.Value);
+      var dockerImage = shouldUseDocker ? kvp.Value : "";
+
+      return new ExecutionRuntime (kvp.Key, kvp.Value, shouldUseDocker, dockerImage);
+    }
+
+    private bool IsEnforcedLocalMachineAndHasDockerImage (IEnumerable<string> configurationItems, out string dockerImage)
```
iam-mholle, line 161: Shouldn't this be either `TryGetDockerImage` or `TryGetDockerImageOfEnforcedLocalMachine` to match the 'pattern'?

BuildScript/BuildTasks/MetadataValueDictionary.cs
```diff
@@ -67,5 +67,22 @@ public bool ContainsKey (string key)
...
       return _rawDictionary.ContainsKey (key);
     }
+
+    public IEnumerable<KeyValuePair<string, string>> GetAllMatches (IEnumerable<string> possibleKeys, Func<Exception> createEmptySequenceException)
+    {
+      var enumeratedPossibleKeys = possibleKeys.ToArray();
+
+      var intersection = _rawDictionary.Keys.Intersect (enumeratedPossibleKeys, _dictionaryStringComparer).ToArray();
+
+      if (intersection.Length == 0)
```
MichaelKetting, line 77: GetAllMatches doens't imply there's an exception possible for no matches. From the usage, there's also no need to unify this into a single operation. => please drop exception logic. In case: Single works differently and has the exception semantics included by the .net framework. Plus, you could not distinguish NULL and NOT FOUND with Single without an exception.

BuildScript/BuildTasks/MetadataValueDictionary.cs
```diff
@@ -67,5 +67,22 @@ public bool ContainsKey (string key)
     {
       return _rawDictionary.ContainsKey (key);
     }
+
+    public IEnumerable<KeyValuePair<string, string>> GetAllMatches (IEnumerable<string> possibleKeys, Func<Exception> createEmptySequenceException)
```
MichaelKetting, line 71: I don't like the name GetAllMatches, but I don't have a better one right now.

BuildScript/BuildTasks/TestConfigurationsFactory.cs
```diff
@@ -144,11 +144,46 @@ private string GetPlatform (IEnumerable<string> configurationItems)
...
+      {
+        return new ExecutionRuntime (MetadataValueConstants.EnforcedLocalMachine, MetadataValueConstants.EnforcedLocalMachine, false, dockerImageOfEnforcedLocalMachine);
+      }
+
+      var kvp = _supportedExecutionRuntimes.Single (enumeratedConfigurationItems, () => CreateMissingConfigurationItemException ("execution runtime"));
 
       var shouldUseDocker = ShouldUseDocker (kvp.Value);
+      var dockerImage = shouldUseDocker ? kvp.Value : "";
+
+      return new ExecutionRuntime (kvp.Key, kvp.Value, shouldUseDocker, dockerImage);
```
MichaelKetting, line 158: We already have a flag, why not always include the docker image?

BuildScript/BuildTasks/TestConfigurationsFactory.cs
```diff
@@ -144,11 +144,46 @@ private string GetPlatform (IEnumerable<string> configurationItems)
 
     private ExecutionRuntime GetExecutionRuntime (IEnumerable<string> configurationItems)
     {
-      var kvp = _supportedExecutionRuntimes.Single (configurationItems, () => CreateMissingConfigurationItemException ("execution runtime"));
+      var enumeratedConfigurationItems = configurationItems.ToArray();
+      if (TryGetDockerImageIfIsEnforcedLocalMachine (enumeratedConfigurationItems, out var dockerImageOfEnforcedLocalMachine))
```
MichaelKetting, line 148: I'm not sure I understand the required complexity here. If we always add the docker image if there is one, could we then not simplify things?

BuildScript/BuildTasks/TestConfigurationsFactory.cs
Stefan-Ilic: I'm not sure I understand what you mean, but If I do: There are cases where no docker image is tangible, like with `LocalMachine` or `EnforcedLocalMachine` without a docker image (which still works because this change is additive)

BuildScript/BuildTasks/TestConfigurationsFactory.cs
Stefan-Ilic: I actually thought doing it like this would reduce complexity. My approach was to not treat `EnforcedLocalMachine` **only** as a flag but additionally allow to provide a docker image along with it. This way I wouldn't have to change existing business logic and architecture and just add one extra branch to `GetExecutionRuntime`'s flow.

If we treated it only as a flag, we'd still need extra handling for `LocalMachine` which never has a docker image associated with it

MichaelKetting: hmm...I'm not convinced this would hurt readability. probably need a pairsession?

BuildScript/BuildTasks/MetadataValueDictionary.cs
Stefan-Ilic: I've renamed it to `Intersect`, I believe it's symmetrical to `Single`

BuildScript/BuildTasks/TestConfigurationsFactory.cs
Stefan-Ilic: Just to be clear: you explicitly want `EnforcedLocalMachine` to only be an extra flag and not an execution runtime on its own like it was until know, did I understand that correctly?

BuildScript/BuildTasks/TestConfigurationsFactory.cs
```diff
@@ -144,11 +138,51 @@ private string GetPlatform (IEnumerable<string> configurationItems)
...
+      }
+      else if (intersection.Length == 1)
+      {
+        return CreateExecutionRuntimeFromSingle (intersection.Single());
+      }
+      else
+      {
+        return CreateExecutionRuntimeFromMultiple (intersection, enumeratedConfigurationItems);
+      }
+    }
```
MichaelKetting, line 156: how about:
var intersection = ...ToList(); // instead of ToArray()
bool hasLocalMachine = intersection.Remove (LocalMachine)
bool hasEnforcedLocalMachine = intersection.Remove (EnforcedlocalMachine)
if (intersection.Count > 1) throw InvalidOpEx ("multiple Docker images found")
var dockerImage = interestion.SingleOrDefault();

then do the IF/ELSEIF/ELSE

Stefan-Ilic: I tried my best to reduce clutter in the method by providing some overloads and extension methods, please take a look if it is still too cluttered or maybe even _okay_

BuildScript/BuildTasks/ExecutionRuntime.cs
```diff
@@ -23,12 +23,22 @@ public class ExecutionRuntime
...
-    public ExecutionRuntime (string key, string value, bool useDocker)
+    public ExecutionRuntime (string key, string value, bool useDocker, string dockerImage)
     {
       Key = key;
       Value = value;
       UseDocker = useDocker;
+      DockerImage = dockerImage;
+    }
+
+    public ExecutionRuntime (string sameKeyAndValue, bool useDocker, string dockerImage)
```
MichaelKetting, line 36: I think we can drop that overload. it's not really all that helpful on the callsite. also, when doing somehting like this, delgate to the same ctor.

PR #18, opened 2020-06-17
RMSRCBUILD-128: Explicitness of tests is ignored in the presence of a filter with adding multiple negated categories when using NUnit2

Stefan-Ilic:
https://re-motion.atlassian.net/browse/RMSRCBUILD-128


Stefan-Ilic:
> 
> 
> Interesting bug.
> 
> Your fix is working fine, but the explicit tests do not show in the `Warnings`, `Inconclusive` or `Skipped` count. Is this intended? Was it like that before? Can't remember.

This might be another "bug" or "behavior" of the NUnit v2 driver. NUnit3 tests + NUnit3 runner show that correctly

PR #19, opened 2021-01-14
RMSRCBUILD-140 Build script helper function Parse-Version-From-ReleaseBranch incorrectly splits strings

raffaelfoidl:
https://re-motion.atlassian.net/browse/RMSRCBUILD-140

### DIFF
--- a/ReleaseProcessScript/tools/Core/main_helper_functions.ps1
+++ b/ReleaseProcessScript/tools/Core/main_helper_functions.ps1
@@ -1,6 +1,6 @@
 function Parse-Version-From-ReleaseBranch ($Branchname)
 {
-  $SplitBranchname = $Branchname.Split("/v")
+  $SplitBranchname = $Branchname.Split(@("/v"), [System.StringSplitOptions]::None)
 
   if ($SplitBranchname.Length -ne 2)
   {
--- a/ReleaseProcessScript/tools/PesterTester/main_helper_functions.Tests.ps1
+++ b/ReleaseProcessScript/tools/PesterTester/main_helper_functions.Tests.ps1
@@ -1,0 +1,10 @@
+. $PSScriptRoot"\..\Core\main_helper_functions.ps1"
+
+Describe "main_helper_functions" {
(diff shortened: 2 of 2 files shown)


raffaelfoidl:
The described problem seems to be related to the execution environment since `.NET Core 2.0` there is a `String.Split(String)` overload, which could explain why it worked sometimes/for someone. With the proposed change to use the `String.Split(String[], StringSplitOptions)` overload, it should work either way.

MichaelKetting:
> The described problem seems to be related to the execution environment since starting with `.NET Core 2.0`, there is a `String.Split(String)` overload, which could explain why it worked sometimes/for someone. With the proposed change to use the `String.Split(String[], StringSplitOptions)` overload, it should work either way.

Nice find. Didn't think of that one.

Stefan-Ilic (approved):
It looks like the tests run on old Powershell 5 now too, so that must have been the error. Very nice job 👍

MichaelKetting (commented):
I believe this bug was introduced after @Stefan-Ilic fround a bug in the split-logic when upgrading the original test suite. Was the test properly red before the fix?

raffaelfoidl:
On PS 7.1, the test is green even before the fix. When running on 5.1, there apparently is made an implicit conversion from `"/v"` to `['/', 'v']`. This observation stems from the fact that, when inspecting `$SplitBranchname` with the debugger (input `"release/v1.2.3"`), its value is `["release", "", "1.2.3"]`. This then leads to an exception `Current branch name is not in a valid format (e. g. release/v1.2.3)` being thrown because `SplitBranchname` is not exactly of length 2. So yes, the new test was red before the fix (on PS 5.1).

MichaelKetting:
@raffaelfoidl Said differently: we cannot ensure that this becomes invalid again but remain green if someone where to change the string.split call. For the test to become read, we would have to force the execution to be done with the powerhsell version used in the production scenario. We really should go and do this whole thing in c# ;)

MichaelKetting (approved):

PR #20, opened 2021-01-15
RMSRCBUILD-147 Wrong version options when releasing a hotfix version

raffaelfoidl:
https://re-motion.atlassian.net/browse/RMSRCBUILD-147

### DIFF
--- a/ReleaseProcessScript/tools/Core/semver_functions.ps1
+++ b/ReleaseProcessScript/tools/Core/semver_functions.ps1
@@ -107,7 +107,7 @@ function Get-Possible-Versions-Hotfix ($Version, $MeantForCurrentVersion = $fals
     $NextPossibleReleaseVersion = "$($CurrentMajor).$($CurrentMinor).$($NextPatch)"  
 
     $NextPreReleaseNumber = [string](1 + $CurrentPreReleaseNumber)
-    $NextPossiblePreReleaseVersion = "$($CurrentMajor).$($CurrentMinor).$($NextPatch)-$($CurrentPreReleaseType).$($NextPreReleaseNumber)" 
+    $NextPossiblePreReleaseVersion = "$($CurrentMajor).$($CurrentMinor).$($CurrentPatch)-$($CurrentPreReleaseType).$($NextPreReleaseNumber)" 
 
     if ($CurrentPreReleaseType -eq "alpha")
     {
--- a/ReleaseProcessScript/tools/Core/semver_functions.ps1
+++ b/ReleaseProcessScript/tools/Core/semver_functions.ps1
@@ -85,44 +85,48 @@ function Get-Possible-Versions-Hotfix ($Version, $MeantForCurrentVersion = $fals
 {
   $Match = Parse-Semver $Version
 
-  $Major = $Match.Groups["major"].ToString()
-  $Minor = $Match.Groups["minor"].ToString()
-  $Patch = $Match.Groups["patch"].ToString()
+  $CurrentMajor = $Match.Groups["major"].ToString()
+  $CurrentMinor = $Match.Groups["minor"].ToString()
+  $CurrentPatch = $Match.Groups["patch"].ToString()
 
   if (-not $MeantForCurrentVersion)
   {
-    $Patch = [string](1 + $Patch)
+    $NextPatch = [string](1 + $CurrentPatch)
   }
-  $NextPossiblePatch = "$($Major).$($Minor).$($Patch)"
+  else 
+  {
+    $NextPatch = $CurrentPatch
+  }
+  $NextPossiblePatchVersion = "$($CurrentMajor).$($CurrentMinor).$($NextPatch)"
 
   #Compute 1.2.3-alpha.4
   if ($Match.Groups["pre"].Success)
   {
-    $Pre = $Match.Groups["pre"].ToString()
-    $PreVersion = $Match.Groups["preversion"].ToString()
-    $NextPossibleFullVersion = "$($Major).$($Minor).$($Patch)"  
+    $CurrentPreReleaseType = $Match.Groups["pre"].ToString()
+    $CurrentPreReleaseNumber = $Match.Groups["preversion"].ToString()
+    $NextPossibleReleaseVersion = "$($CurrentMajor).$($CurrentMinor).$($NextPatch)"  
 
-    $NextPreVersion = [string](1 + $PreVersion)
-    $NextPossiblePreVersion = "$($Major).$($Minor).$($Patch)-$($Pre).$($NextPreVersion)" 
+    $NextPreReleaseNumber = [string](1 + $CurrentPreReleaseNumber)
+    $NextPossiblePreReleaseVersion = "$($CurrentMajor).$($CurrentMinor).$($NextPatch)-$($CurrentPreReleaseType).$($NextPreReleaseNumber)" 
 
-    if ($Pre -eq "alpha")
+    if ($CurrentPreReleaseType -eq "alpha")
     {
-      $NextPossiblePre = "$($Major).$($Minor).$($Patch)-beta.1"
+      $NextPossibleBetaVersion = "$($CurrentMajor).$($CurrentMinor).$($CurrentPatch)-beta.1"
 
-      return $NextPossiblePreVersion, $NextPossiblePre, $NextPossibleFullVersion
+      return $NextPossiblePreReleaseVersion, $NextPossibleBetaVersion, $NextPossibleReleaseVersion
     }
-    elseif ($Pre -eq "beta")
+    elseif ($CurrentPreReleaseType -eq "beta")
     {
-      return $NextPossiblePreVersion, $NextPossibleFullVersion
+      return $NextPossiblePreReleaseVersion, $NextPossibleReleaseVersion
     }
-    elseif ($Pre -eq "rc")
+    elseif ($CurrentPreReleaseType -eq "rc")
     {
-      return $NextPossiblePreVersion, $NextPossibleFullVersion
+      return $NextPossiblePreReleaseVersion, $NextPossibleReleaseVersion
     }
   }
   else
   {
-    return $NextPossiblePatch, "$($NextPossiblePatch)-alpha.1", "$($NextPossiblePatch)-beta.1"
+    return $NextPossiblePatchVersion, "$($NextPossiblePatchVersion)-alpha.1", "$($NextPossiblePatchVersion)-beta.1"
   } 
 }
 
--- a/ReleaseProcessScript/tools/PesterTester/semver_functions.Tests.ps1
+++ b/ReleaseProcessScript/tools/PesterTester/semver_functions.Tests.ps1
@@ -95,6 +95,34 @@ Describe "semver_functions" {
 
       Get-Possible-Versions-Hotfix $Version | Should Be $NextVersions
     }
+
+    It "Get-Possible-Next-Versions-Hotfix_WithAlpha_NotMeantForCurrentVersion_ShouldReturnArray" {
+      $Version = "2.28.0-alpha.1"
+      $NextVersions = "2.28.0-alpha.2", "2.28.0-beta.1", "2.28.1"
+
+      Get-Possible-Versions-Hotfix $Version | Should Be $NextVersions
+    }
+
+    It "Get-Possible-Next-Versions-Hotfix_WithBeta_NotMeantForCurrentVersion_ShouldReturnArray" {
+      $Version = "2.28.0-beta.2"
+      $NextVersions = "2.28.0-beta.3", "2.28.1"
+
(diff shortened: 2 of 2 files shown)


ReleaseProcessScript/tools/PesterTester/semver_functions.Tests.ps1
```diff
@@ -95,6 +95,34 @@ Describe "semver_functions" {
 
       Get-Possible-Versions-Hotfix $Version | Should Be $NextVersions
     }
+
+    It "Get-Possible-Next-Versions-Hotfix_WithAlpha_NotMeantForCurrentVersion_ShouldReturnArray" {
+      $Version = "2.28.0-alpha.1"
+      $NextVersions = "2.28.0-alpha.2", "2.28.0-beta.1", "2.28.1"
```
iam-mholle, line 101: Shouldn't `2.28.0-rc.1` also be part of this? `Get-Possible-Next-Versions-Hotfix_WithRc_NotMeantForCurrentVersion_ShouldReturnArray` shows the possibility of `rc`'s on hotfix branches (I guess?). @MichaelKetting ?

MichaelKetting: that would be a separate issue. 

As to the reasoning about rc pro/contra: we should never want to start an rc with open issues to be fixed in the rc. For practical purposes, it might be helpful, on occasion. I still hope it's uncommon enough a scenario that it won't matter so there's no need to add an issue for this.

ReleaseProcessScript/tools/Core/semver_functions.ps1
```diff
@@ -103,11 +103,12 @@ function Get-Possible-Versions-Hotfix ($Version, $MeantForCurrentVersion = $fals
     $NextPossibleFullVersion = "$($Major).$($Minor).$($Patch)"  
 
     $NextPreVersion = [string](1 + $PreVersion)
-    $NextPossiblePreVersion = "$($Major).$($Minor).$($Patch)-$($Pre).$($NextPreVersion)" 
+    $NextPrePatch = $Match.Groups["patch"].ToString()
```
iam-mholle, line 106: I think it would be easier to understand if `$Patch` of line 90 would be immutable and a new variable is created for the possibly mutated patch number on lines 92-96. This way you could just use the initial `$Patch` here without extracting it from the match a second time.

iam-mholle (changes requested):

Stefan-Ilic (approved):
I actually agree with Marcel but I understand that this very small bug fix doesn't need to restructure the entire method if we don't have to. Battle it out with him 👍

ReleaseProcessScript/tools/Core/semver_functions.ps1
MichaelKetting: I'm with @iam-mholle on this one. The mix of mutable and mixed naming in this method is not helpful towards maintainablity. I tried to come up with a compromise, but there just isn't one. 

Each variable should be assigned only once and b) have a clear name. It will be okay if we duplicate the building of the version number at this time since there's less business logic in the format *) than in using the correct number-parts in the first place.

*) Actually, the version format is completely fixed and easily reviewed so there's no worry this could break in the future. In theory, we would also have a version builder and then only do the formatting once, but that's nice to have, nothing more.

ReleaseProcessScript/tools/PesterTester/semver_functions.Tests.ps1
raffaelfoidl: I have refactored the method such that it (hopefully) has a better naming and the initially extracted version part variables (from the regex) are not mutated.

ReleaseProcessScript/tools/Core/semver_functions.ps1
```diff
@@ -85,44 +85,48 @@ function Get-Possible-Versions-Hotfix ($Version, $MeantForCurrentVersion = $fals
...
-    $NextPreVersion = [string](1 + $PreVersion)
-    $NextPossiblePreVersion = "$($Major).$($Minor).$($Patch)-$($Pre).$($NextPreVersion)" 
+    $NextPreReleaseNumber = [string](1 + $OriginalPreReleaseNumber)
+    $NextPossiblePreReleaseVersion = "$($OriginalMajor).$($OriginalMinor).$($OriginalPatch)-$($PreReleaseType).$($NextPreReleaseNumber)" 
 
-    if ($Pre -eq "alpha")
+    if ($PreReleaseType -eq "alpha")
     {
-      $NextPossiblePre = "$($Major).$($Minor).$($Patch)-beta.1"
+      $NextPossibleBetaVersion = "$($OriginalMajor).$($OriginalMinor).$($OriginalPatch)-beta.1"
```
iam-mholle, line 114: I would prefer `Current...` over `Original...`. Other than that, it seems fine.

iam-mholle (approved):

ReleaseProcessScript/tools/Core/semver_functions.ps1
raffaelfoidl: I have refactored the method such that it (hopefully) has a better naming scheme and the initially extracted version part variables (from the regex) are not mutated.

ReleaseProcessScript/tools/Core/semver_functions.ps1
raffaelfoidl: Ok, I'm just waiting for @MichaelKetting's general opinion before renaming them so that I don't have to change the PR too often.

ReleaseProcessScript/tools/Core/semver_functions.ps1
```diff
@@ -85,44 +85,48 @@ function Get-Possible-Versions-Hotfix ($Version, $MeantForCurrentVersion = $fals
...
+  }
+  $NextPossiblePatchVersion = "$($OriginalMajor).$($OriginalMinor).$($NextPatch)"
 
   #Compute 1.2.3-alpha.4
   if ($Match.Groups["pre"].Success)
   {
-    $Pre = $Match.Groups["pre"].ToString()
-    $PreVersion = $Match.Groups["preversion"].ToString()
-    $NextPossibleFullVersion = "$($Major).$($Minor).$($Patch)"  
+    $PreReleaseType = $Match.Groups["pre"].ToString()
```
MichaelKetting, line 105: CurrentPreReleaseType

ReleaseProcessScript/tools/Core/semver_functions.ps1
```diff
@@ -85,44 +85,48 @@ function Get-Possible-Versions-Hotfix ($Version, $MeantForCurrentVersion = $fals
 {
   $Match = Parse-Semver $Version
 
-  $Major = $Match.Groups["major"].ToString()
-  $Minor = $Match.Groups["minor"].ToString()
-  $Patch = $Match.Groups["patch"].ToString()
+  $OriginalMajor = $Match.Groups["major"].ToString()
```
MichaelKetting, line 88: Current...

ReleaseProcessScript/tools/Core/semver_functions.ps1
```diff
@@ -85,44 +85,48 @@ function Get-Possible-Versions-Hotfix ($Version, $MeantForCurrentVersion = $fals
...
 
   #Compute 1.2.3-alpha.4
   if ($Match.Groups["pre"].Success)
   {
-    $Pre = $Match.Groups["pre"].ToString()
-    $PreVersion = $Match.Groups["preversion"].ToString()
-    $NextPossibleFullVersion = "$($Major).$($Minor).$($Patch)"  
+    $PreReleaseType = $Match.Groups["pre"].ToString()
+    $OriginalPreReleaseNumber = $Match.Groups["preversion"].ToString()
+    $NextPossibleFullVersion = "$($OriginalMajor).$($OriginalMinor).$($NextPatch)"
```
MichaelKetting, line 107: ...ReleaseVersion instead of FullVersion

ReleaseProcessScript/tools/Core/semver_functions.ps1
```diff
@@ -107,7 +107,7 @@ function Get-Possible-Versions-Hotfix ($Version, $MeantForCurrentVersion = $fals
     $NextPossibleFullVersion = "$($OriginalMajor).$($OriginalMinor).$($NextPatch)"  
 
     $NextPreReleaseNumber = [string](1 + $OriginalPreReleaseNumber)
-    $NextPossiblePreReleaseVersion = "$($OriginalMajor).$($OriginalMinor).$($NextPatch)-$($PreReleaseType).$($NextPreReleaseNumber)" 
+    $NextPossiblePreReleaseVersion = "$($OriginalMajor).$($OriginalMinor).$($OriginalPatch)-$($PreReleaseType).$($NextPreReleaseNumber)"
```
MichaelKetting, line 110: shouldn't there be a test for this in the second commit?

MichaelKetting (changes requested):

ReleaseProcessScript/tools/Core/semver_functions.ps1
raffaelfoidl: The added tests are in the first commit. Do you want me to shift them to the second one?

MichaelKetting: yes, please. the first commit was just refactoring, so there shouldn't be behavioral changes and all tests green. The added test can't be green without the fix, so it belongs to the second commit. alternatively, you could add the test in the test in the first commit with the wrong list of versions and just fix the expectation in the second commit.

MichaelKetting (commented):
I'm not sure about the final outcome, but didn't we discuss the option of adding the tests with the wrong expections to the first commit and then just fixing the tests and the production code in the second commit?

raffaelfoidl:
> I'm not sure about the final outcome, but didn't we discuss the option of adding the tests with the wrong expections to the first commit and then just fixing the tests and the production code in the second commit?

Yes, you mentioned it. As far as I understood, it was just one of the two options (either adding the tests in the second commit or adding them in the first and fixing them in the second one). I chose the first alternative since it was quicker to implement. However, if you prefer the approach of fixing the test expectations along with the production code in the second test, I can change it.

(The comment I am referring to: https://github.com/re-motion/SharedSource-Build/pull/20#discussion_r562578599)

MichaelKetting:
> 
> 
> > I'm not sure about the final outcome, but didn't we discuss the option of adding the tests with the wrong expections to the first commit and then just fixing the tests and the production code in the second commit?
> 
> Yes, you mentioned it. As far as I understood, it was just one of the two options (either adding the tests in the second commit or adding them in the first and fixing them in the second one). I chose the first alternative since it was quicker to implement. However, if you prefer the approach of fixing the test expectations along with the production code in the second test, I can change it.
> 
> (The comment I am referring to: [#20 (comment)](https://github.com/re-motion/SharedSource-Build/pull/20#discussion_r562578599))

@iam-mholle & @Stefan-Ilic  what do you think?

iam-mholle:
> > > I'm not sure about the final outcome, but didn't we discuss the option of adding the tests with the wrong expections to the first commit and then just fixing the tests and the production code in the second commit?
> > 
> > 
> > Yes, you mentioned it. As far as I understood, it was just one of the two options (either adding the tests in the second commit or adding them in the first and fixing them in the second one). I chose the first alternative since it was quicker to implement. However, if you prefer the approach of fixing the test expectations along with the production code in the second test, I can change it.
> > (The comment I am referring to: [#20 (comment)](https://github.com/re-motion/SharedSource-Build/pull/20#discussion_r562578599))
> 
> @iam-mholle & @Stefan-Ilic what do you think?

IMO tests and the related implementation belong together commit-wise. The first commit is fine as-is, as it is a refactoring. The second commit adds a test and the related fix. As this is no change in requirements, I don't see any value in separating the test from the fix in any form. Hope this makes sense somehow.

MichaelKetting (approved):

Stefan-Ilic:
> > > I'm not sure about the final outcome, but didn't we discuss the option of adding the tests with the wrong expections to the first commit and then just fixing the tests and the production code in the second commit?
> > 
> > 
> > Yes, you mentioned it. As far as I understood, it was just one of the two options (either adding the tests in the second commit or adding them in the first and fixing them in the second one). I chose the first alternative since it was quicker to implement. However, if you prefer the approach of fixing the test expectations along with the production code in the second test, I can change it.
> > (The comment I am referring to: [#20 (comment)](https://github.com/re-motion/SharedSource-Build/pull/20#discussion_r562578599))
> 
> @iam-mholle & @Stefan-Ilic what do you think?

To me, the most important thing is that in a bugfix commit/PR there is also a change in the tests, which we'd have with both alternatives. Having the test in the first commit, its greatest use to me would be to make sure the refactoring didn't change any functionality, but since we're reasonably sure about that and since the tests weren't present before anyway (what's a commit sooner or later) and since we already have this state as it is I wouldn't change anything anymore

PR #21, opened 2021-02-04
RMSRCBUILD-146 At the end of the hotfix release, the DevelopmentForNextRelease script is not executed

raffaelfoidl:
https://re-motion.atlassian.net/browse/RMSRCBUILD-146

### DIFF
--- a/ReleaseProcessScript/tools/Core/main_functions.ps1
+++ b/ReleaseProcessScript/tools/Core/main_functions.ps1
@@ -227,7 +227,7 @@ function Release-On-Master ()
   git checkout "develop" 2>&1 | Write-Host
 
 
-  #develop should be prepared with the NextVersion, as the develop code should now be filled with the next Version number
+  #develop should be prepared with the NextVersion, as the develop code should now be filled with the next version number
   Invoke-MsBuild-And-Commit -Version $NextVersion -MsBuildMode "developmentForNextRelease"
      
   git checkout $ReleaseBranchname --quiet
@@ -461,7 +461,12 @@ function Continue-Patch-Release ()
   git tag -a $Tagname -m $Tagname 2>&1
 
   $NextPatchVersion = Get-Next-Patch $CurrentVersion
-  git branch "hotfix/v$($NextPatchVersion)"
+  git checkout -b "hotfix/v$($NextPatchVersion)" 2>&1 | Write-Host
+
+  #hotfix branch should be prepared with the NextPatchVersion, as the hotfix branch code should now be filled with the next version number
+  Invoke-MsBuild-And-Commit -Version $NextPatchVersion -MsBuildMode "developmentForNextRelease"
+
+  git checkout $MergeTargetBranchName --quiet
 
   if ($DoNotPush)
   {


raffaelfoidl:
Together with @Stefan-Ilic, I have tested different configurations.

**Without the changes proposed in this PR**:  
* releasing from `develop`: correct behaviour could be observed
* releasing a hotfix: `DevelopmentForNextRelease` was not executed

**With the changes proposed in this PR**:
* releasing a hotfix: `DevelopmentForNextRelease` was executed - more specifically, the MSBuild target `UpdateAssemblyInfosForDevelopment` (specified in `Build\Customizations\ReleaseProcessScript.config`) was executed

Thus, we concluded that it now works as intended.

ReleaseProcessScript/tools/Core/main_functions.ps1
```diff
@@ -461,7 +461,12 @@ function Continue-Patch-Release ()
   git tag -a $Tagname -m $Tagname 2>&1
 
   $NextPatchVersion = Get-Next-Patch $CurrentVersion
-  git branch "hotfix/v$($NextPatchVersion)"
+  git checkout -b "hotfix/v$($NextPatchVersion)"
```
iam-mholle, line 464: What's the reason behind not suppressing the `Switched to a new branch 'xyz'` message here?

raffaelfoidl: There is no such reason in particular. However, the other occurrence of `developmentForNextRelease`, the message is not suppressed, either.

https://github.com/re-motion/SharedSource-Build/blob/992d341b272f55260b520cc2fd5307b5e5301006/ReleaseProcessScript/tools/Core/main_functions.ps1#L226-L231

I cannot tell if this is was done on purpose. However, for the sake of consistency and in order to implement your suggestion, what about the following:

`git checkout -q -b "hotfix/v$($NextPatchVersion)" 2>&1 | Write-Host`

This would suppress the feedback message you mentioned, but errors like "branch already exists" should still be displayed (which is appreciated, I guess?).

iam-mholle: I do not know if the output could be informative for the user, might be. In this case, `git checkout -b "hotfix/v$($NextPatchVersion)" 2>&1 | Write-Host` (note that there is no `-q`) would be the thing to do. Else, just use `git checkout -q -b "hotfix/v$($NextPatchVersion)"` or `git checkout -q -b "hotfix/v$($NextPatchVersion)" 2>&1 | Write-Host` if the stream redirect is really needed.

iam-mholle (approved):

ReleaseProcessScript/tools/Core/main_functions.ps1
```diff
@@ -461,7 +461,12 @@ function Continue-Patch-Release ()
...
+
+  #hotfix branch should be prepared with the NextPatchVersion, as the hotfix branch code should now be filled with the next Version number
```
MichaelKetting, line 466: there's a case typo in "Version number"

ReleaseProcessScript/tools/Core/main_functions.ps1
MichaelKetting: It looks like this has not yet been incorporated?

MichaelKetting (changes requested):

ReleaseProcessScript/tools/Core/main_functions.ps1
raffaelfoidl: You are right. I was curious as to what you think because I don't have strong opinions on any of the possibilities presented. I cannot tell if the stream redirection is really needed, however, I'd personally refrain from introducing inconsistencies, especially since I do not know the motivation behind invoking git the way it was done in the snippet linked above.

a) `git checkout -b "hotfix/v$($NextPatchVersion)"` (initial proposal)
b) `git checkout -q -b "hotfix/v$($NextPatchVersion)"`

c) `git checkout -b "hotfix/v$($NextPatchVersion)" 2>&1 | Write-Host` (current version)
d) `git checkout -q -b "hotfix/v$($NextPatchVersion)" 2>&1 | Write-Host`

However, if I had to choose, I'd go for option c (I have updated the PR accordingly). The message could be informative for the user such that they are able to trace the release process better. In my opinion, the single log message `Switched to a new branch 'hotfix/vX.Y.Z'` is not disruptive, hence I don't have a strong desire to suppress it.

MichaelKetting: Completely valid reasoning. Would also have been my choice and reasoning.

MichaelKetting (approved):

PR #22, opened 2021-02-12
RMSRCBUILD-151: Calculation of Next Prerelease Version Does Not Work for Hotfix Branches

raffaelfoidl:
https://re-motion.atlassian.net/browse/RMSRCBUILD-151


ReleaseProcessScript/tools/Core/git_base_functions.ps1
```diff
@@ -22,14 +22,29 @@ function Get-Current-Branchname ()
...
+{
+  #pre-release versions are sorted in order of specification
+  $gitParams = @("-c", "versionsort.suffix=-alpha", "-c", "versionsort.suffix=-beta", "-c", "versionsort.suffix=-rc")
+  $refParams = @("refs/tags", "--sort=-version:refname", "--format=`"%(refname:short)`"", "--count=1")
+  
+  #version tags are looked for starting from revision $From (backwards in history) until $To (empty string is equivalent to "unbounded")
+  $refParams += If ([string]::IsNullOrEmpty($From)) {"--merged=HEAD"} Else {"--merged=`"$From`""}
+  $refParams += If ([string]::IsNullOrEmpty($To)) {""} Else {"--contains=`"$To`""}
+
+  return @($gitParams, $refParams)
```
raffaelfoidl, line 47: This is a noteworthy change I would like to spend some words on. The previous approach proved unreliable in certain scenarios because tags that were not merged into the current branch (e. g. from a support branch) were considered "closer" than tags from a merged branch, despite them being more recent (e. g. from a prerelease branch).

The example below demonstrates the issue. It might be a bit contrived, but hopefully shows the essence of the problem.

Assume the following repository (I omitted the hotfix/v2.28.1 branch for the sake of brevity):
```
PS> git log --all --decorate --oneline --graph
*   fdd2b09 (HEAD -> hotfix/v2.28.2) Merge branch 'prelease/v2.28.2-alpha.1' into hotfix/v2.28.2
|\
| * 5b8e87b (tag: v2.28.2-alpha.1, prelease/v2.28.2-alpha.1) Update metadata to version '2.28.2-alpha.1'.
|/
* 97935fa Feature-C
* 7dd3a53 (tag: v2.28.1, support/v2.28) Feature-B
* 31777be Feature-A
* a7360b4 (tag: v2.28.0, master) Initial Release
```

The previous approach returns the tag from commit "Feature-B". However, `v2.28.2-alpha.1` would be correct.
```
PS> git describe --match "v[0-9]*" --abbrev=0
v2.28.1
```

The proposed solution relies less on the DAG structure of the git log (as it did until now), but more on version numbers that are strictly monotonically increasing. The `--sort=-version:refname` sorts all version tags such that the newest version is the first line in the output. This assumption is hopefully a reasonable compromise.

```
PS> git -c "versionsort.suffix=-alpha" -c "versionsort.suffix=-beta" -c "versionsort.suffix=-rc" -c "versionsort.suffix=`"`"" for-each-ref refs/tags --sort=-version:refname --format="%(refname:short)" --count=1
v2.28.2-alpha.1
```

MichaelKetting: Hmm... I just tried to understand this and google / stackoverflow suggests:
`git describe`

And it does seem to work and print out the correct tags. tested it in hotfix/2.28 and on develop, but we should do more testing.

https://stackoverflow.com/questions/33275853/git-find-all-tags-reachable-from-a-commit
https://git-scm.com/docs/git-describe
> The command finds the most recent tag that is reachable from a commit

Could you please check if this is a simpler/better solution which also gives us the correct results? Disclaimer: you may have to do a string-split on the result because it prints
> v2.28.0-alpha.3-3-g3a9b7c46

But there could very well be a parameter to make this all better out of the box.

raffaelfoidl: Right at the start: Sorry for the lengthy answer.

Yes, I was able to to retrieve the correct tag in re-motion as well - after all, the function has been there in this form for quite some time. It's just that, when testing with different combinations of branches and tags (e. g. do not fall back to a tag from the support branch or even master), I noticed potentially unwanted behaviour.

For example, take the log depicted in the screenshot below (see [git commands](https://github.com/re-motion/SharedSource-Build/blob/1ef9a8d05a6eda22c12a8c84c878fbcc7e723758/ReleaseProcessScript/tools/PesterTester/GitCommandsForReferenceDir/Hotfix-PreRelease-Beta/Initialize.ps1) from the respective tests; "First commit", "Second commit" including 1.0.0 and 1.1.0 tags are created by the infrastructure beforehand).
image

The `Get-Last-Version-Of-Branch-From-Tag` should return `v2.28.1-beta.1` in this case, exemplified by the corresponding test.
https://github.com/re-motion/SharedSource-Build/blob/1ef9a8d05a6eda22c12a8c84c878fbcc7e723758/ReleaseProcessScript/tools/PesterTester/IntegrationTests.Tests.ps1#L80-L86

However, I was often (not always, that's the great thing I don't understand) greeted with a test failure because the function returned the `v1.1.0` tag.

```
PS> git describe --match "v[0-9]*" --abbrev=0
v1.1.0

PS> git describe --match "v[0-9]*" --abbrev=0 --debug
describe HEAD
No exact match on refs or tags, searching to describe
finished search at c6442cf2e010806e33d20243e83b1ef5f51e7c6f
 annotated          4 v1.1.0
 annotated          5 v2.28.1-beta.1
 annotated          5 v1.0.0
traversed 6 commits
v1.1.0
```

image

From what I was able to gather, the numeric values printed out with the `--debug` flag (and if you leave out ``--abrev=0``) is the "distance", i. e. the number of commits from HEAD that were traversed - in some kind of depth-first search way. Starting from the latest commit, it's four hops to the v1.1.0 tag (then backtracking) and another one to the v2.28.1-beta.1 tag. However, that was just an assumption.

The "nice" thing is that I was not able to find information that fully describe the algorithm employed by git. The behaviour above is not consistent to a degree of 100 %. There is a failed Shared Source Build build of this 151-branch where the test was successful - however, an analogous test with the a very similar repo structure failed. It was an alpha.2 release with a previous alpha.1 release and the function returned alpha.1 as latest version instead of alpha.2, which is even more strange in my opinion - see screenshot below for the repository.

image

In summary, I just don't fully understand, which made me uncomfortable not to change anything. Since I was not able to find explanations for this inconsistent behaviour (different outputs for the very same repository on different executions), I took it as the function being not as reliable as we might want it to be as it relies on measuring "closeness" in a DAG, which can be tricky and may have different notions depending on what to achieve. That's why the alternative I proposed solely relies on the order relation that's defined for semver versions - in other words we have a more lengthy command, but more predictable results.

If such "problematic" cases are not relevant in practice, I can gladly retract the changes and stick to the old solution. I am not even sure if this undesirable behaviour only occurs in repositories as small as in the test cases I created locally - "real" repositories have quite different distances between version tags.

Apologies again for my extensive answer. Hopefully, I was able to explain my line of thinking in a digestible, linear way. Nevertheless, further discussion might be better lead in person.

MichaelKetting: @raffaelfoidl 
Did you do annotated tags or just tags?

And yes, abbrev is weird. Here's the docs
> Instead of using the default 7 hexadecimal digits as the abbreviated object name, use <n> digits, or as many digits as needed to form a unique object name. **An <n> of 0 will suppress long format, only showing the closest tag.**

So, `abbrev=0` seems to be what we need. 

Yes, a live session would be helpful.

raffaelfoidl: I used annotated tags.

MichaelKetting: annotated: okay, that's not it, then.
Another question:
> The command finds the most recent tag that is reachable from a commit. 

The docs refer to the "most recent tag".  Is it possible that you created the tags in the wrong temporal order?

raffaelfoidl: Unfortunately, the temporal order should be correct (from the command listings and the log):
```
v1.0.0:                10:04:21
v1.1.0:                10:04:21
v2.28.1-alpha.1: 10:04:23
v2.28.1-alpha.2: 10:04:23
```

Funny enough, if I add `--candidates=1` to the command, it seems to return the correct result.
> Instead of considering only the 10 most recent tags as candidates to describe the input commit-ish consider up to <n> candidates. Increasing <n> above 10 will take slightly longer but may produce a more accurate result. An <n> of 0 will cause only exact matches to be output.

```
PS> git describe --abbrev=0 --debug
describe HEAD
No exact match on refs or tags, searching to describe
finished search at ae05250e1557b88455b9db08289ce15ce09ac1a9
 annotated          7 v1.1.0
 annotated          8 v2.28.1-alpha.2
 annotated          8 v2.28.1-alpha.1
 annotated          8 v1.0.0
traversed 9 commits
v1.1.0

PS> git describe --candidates=1 --abbrev=0 --debug
describe HEAD
No exact match on refs or tags, searching to describe
 annotated          8 v2.28.1-alpha.2
traversed 9 commits
more than 1 tags found; listed 1 most recent
gave up search at 07896fb7b383d61a015f8390b11eaf3634b1ec25
v2.28.1-alpha.2
```

With this addition, the tests are green, even for many executions. I am a bit surprised that the revision after `gave up search` is not the one with the output-tag, but what would have been the second recent tag (`v2.28.1-alpha.1`).

So even though that seems to work, I am anxious as I do not fully understand what is going on. Anyway, with `git describe` I would also still have to figure out how stop the search at the point from which the hotfix branch started (i. e. tags from support/master should not even be an option, for the case that there is no tag on the hotfix branch yet).

MichaelKetting: About stopping: we could just use pattern logic here and match all tags that fit the branch name's version. that way, we could just look at all the tags and would still get the correct result.

One other option about the "recent" maybe there's a timestamp issue, i.e. tags created within the same second are ambiguous.

ReleaseProcessScript/tools/Core/git_base_functions.ps1
```diff
@@ -22,14 +22,36 @@ function Get-Current-Branchname ()
   return $(git symbolic-ref --short -q HEAD)
 }
 
-function Get-Last-Version-Of-Branch-From-Tag-Exists ($Branchname)
+function Get-Last-Version-Of-Branch-From-Tag-Exists ($From, $To)
```
Stefan-Ilic, line 25: Could you rename the parameters to make it more explicit what `$From` and `$To` are? Are they tags names, branch names, hashes etc.

ReleaseProcessScript/tools/Core/git_base_functions.ps1
```diff
@@ -22,14 +22,36 @@ function Get-Current-Branchname ()
...
 {
-  return (git describe $Branchname --match "v[0-9]*" --abbrev=0 2>&1) -and $?
+  $params = Get-Parameters-To-Retrieve-Last-Version-From-Tag $From $To
+  return (git $params[0] for-each-ref $params[1] 2>&1) -and $?
 }
 
-function Get-Last-Version-Of-Branch-From-Tag ($Branchname)
+function Get-Last-Version-Of-Branch-From-Tag ($From, $To)
```
Stefan-Ilic, line 31: Same thing with the parameters here, see previous comment about parameter names

ReleaseProcessScript/tools/Core/git_base_functions.ps1
```diff
@@ -22,14 +22,36 @@ function Get-Current-Branchname ()
...
 {
-  return git describe $Branchname --match "v[0-9]*" --abbrev=0
+  $params = Get-Parameters-To-Retrieve-Last-Version-From-Tag $From $To
+  return git $params[0] for-each-ref $params[1]
+}
+
+function Get-Parameters-To-Retrieve-Last-Version-From-Tag ($From, $To)
```
Stefan-Ilic, line 37: Also here

ReleaseProcessScript/tools/Core/git_base_functions.ps1
```diff
@@ -22,14 +22,36 @@ function Get-Current-Branchname ()
...
+function Get-Last-Version-Of-Branch-From-Tag ($From, $To)
 {
-  return git describe $Branchname --match "v[0-9]*" --abbrev=0
+  $params = Get-Parameters-To-Retrieve-Last-Version-From-Tag $From $To
+  return git $params[0] for-each-ref $params[1]
+}
+
+function Get-Parameters-To-Retrieve-Last-Version-From-Tag ($From, $To) 
+{
+  #The latest tag is determined by gathering all tags reachable from $From to $To (backwards in history). The
```
Stefan-Ilic, line 39: And in this comment

Stefan-Ilic (changes requested):

ReleaseProcessScript/tools/Core/git_base_functions.ps1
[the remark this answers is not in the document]
raffaelfoidl: Ah, yes, could have made a note. I just thought a separate ticket or PR would have been overkill.

image

raffaelfoidl: I am no expert on git terminology, so I might need help from someone who is on that one.

`$From` and `$To` are used to denote revisions. They can be tag names (implying the commit the tag points to - I only tested it with annotated tags, though, would have to test lightweight tags as well), branch names (referring to the tip of the branch) or commit hashes (self-explanatory).

The documentation describes the parameter as _object_ and/or commit:

> --merged[=\<object>]
Only list refs whose tips are reachable from the specified commit (HEAD if not specified).

So in the end it is a commit. One might say the parameters accept `commit-ish`, as defined by the official documentation:

> commit-ish (also committish)
A commit object or an object that can be recursively dereferenced to a commit object. The following are all commit-ishes: a commit object, a tag object that points to a commit object, a tag object that points to a tag object that points to a commit object, etc.

(source: https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefcommit-ishacommit-ishalsocommittish)

I am not sure if, by that definition, the name of a branch is also a `commit-ish` - in any case, it definitely works as parameter for this command.

So would `$FromCommitish` and `$ToCommitish` be more to your liking? I personally wouldn't fancy it, but if you think it's more descriptive, I'll gladly change it - the same goes for any other (better) suggestion you might have.

Stefan-Ilic: I agree that `$FromCommitish` sounds kind of dumb, but it **is** the official terminology. The information you collected in your answeris very good, thank you very much for it. Considering your opinion, may I suggest adding the [powershell equivalent of javadoc](https://docs.microsoft.com/en-us/powershell/scripting/developer/help/examples-of-comment-based-help?view=powershell-7.1) to the functions? I genuinely feel like we have to _somehow_ add the information what `$To` and `$From` actually are, maybe it's changing the parameter name, maybe it's using powershell's comment-based-help or maybe it's just adding plain comments underneath the signature. You could try adding the comment-based-help thingies and see if it suits the function/file and looks any good. Otherwise, maybe just 2 plain comment lines briefly explaining what a commit-ish object is would suffice, I will leave it up to you (and finally Michael) to decide what's best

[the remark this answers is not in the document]
Stefan-Ilic: Would you kindly make a new slide with your change and write your name and change date at the top left like I started doing? I know it's a little dumb, but my initial proposal to remake the diagram with draw.io (or any other text-based diagram tool that can be checked into source control) was shot down because of budget reasons, so this "poor man's versioning" is better than nothing imo.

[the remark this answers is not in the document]
raffaelfoidl: I have inserted it as an extra slide and added an additional slide concerning project hotfixes.

raffaelfoidl: I completely agree with you that it's best to include information on what `$To` and `$From` really stand for. I have updated the respective file with "ordinary" comments next to the formal arguments. I actually considered the Comment-Based Help you linked before opening the PR but opted against it because this function/these functions would be "the odd ones out" in the greater scope of the release process script as a whole (such extensive documentation is only found in the main entry points in `ReleaseProcessScript.psm1`). However, one could argue that it's worth adding such extensive documentation, i. e. moving comments from function body to help text and possibly even adding one or two examples in this help text. This should, in my opinion, only apply to this function, though - for the other ones you commented on below, short comments next next to the parameter definition should suffice (to me at least, since I do not want to spam the file with repetetive documentation).

But I agree that, of course, Michael has the final say in what's the best route to go.

Stefan-Ilic (approved):

ReleaseProcessScript/tools/Core/main_helper_functions.ps1
```diff
@@ -49,6 +49,36 @@ function Get-Develop-Current-Version ($StartReleasebranch)
     return Read-Current-Version
   }
 
+  $MostRecentVersion = Get-Develop-Most-Recent-Version
+    
+  $PossibleVersions = Get-Possible-Next-Versions-Develop $MostRecentVersion $WithoutPrerelease
+  Write-Host "Please choose Release Version:"
```
MichaelKetting, line 55: where did we have this message before? And why does it remain there?

ReleaseProcessScript/tools/Core/main_helper_functions.ps1
```diff
@@ -49,6 +49,36 @@ function Get-Develop-Current-Version ($StartReleasebranch)
...
+
+    $PossibleVersions = Get-Possible-Versions-Hotfix $MostRecentVersion $true
+    Write-Host "Please choose Release Version:"
+    $CurrentVersion = Read-Version-Choice $PossibleVersions
+  }
+
+  return $CurrentVersion
+}
+
+function Get-Develop-Most-Recent-Version () {
```
MichaelKetting, line 81: linebreak

ReleaseProcessScript/tools/Core/main_helper_functions.ps1
```diff
@@ -69,27 +99,27 @@ function Get-Develop-Current-Version ($StartReleasebranch)
...
-  $CurrentVersion = Read-Version-Choice $PossibleVersions
 
-  return $CurrentVersion
+  return $MostRecentVersion
 }
 
-function Get-Hotfix-Current-Version ($LastVersion, $StartReleasePhase)
-{
-  if ($StartReleasePhase)
+function Get-Hotfix-Most-Recent-Version () {
```
MichaelKetting, line 106: linebreak

ReleaseProcessScript/tools/Core/git_base_functions.ps1
```diff
@@ -22,14 +22,48 @@ function Get-Current-Branchname ()
   return $(git symbolic-ref --short -q HEAD)
 }
 
-function Get-Last-Version-Of-Branch-From-Tag-Exists ($Branchname)
+function Get-Last-Version-Of-Branch-From-Tag-Exists
+(
+  $From, #A commit-ish from wich to start looking for version tags (backwards in history). Can be a tag name, a branch name, a commit hash, HEAD etc. Default: HEAD
```
MichaelKetting, line 27: I cannot reply to the original discussion so... 

In this particular context, the "FromCommitish" might be overkill and makes if harder to read.  I don't think we have a maintenance issue, leaving the names / strings as they are.

I do not like the fact that there's no whitespace between the comment-hash and the comment text. This just reads strange.

MichaelKetting (changes requested):

ReleaseProcessScript/tools/Core/main_helper_functions.ps1
raffaelfoidl: Before, the message "Please choose Release Version:" was only printed in the function `Get-Develop-Current-Version` (and not `Get-Hotfix-Current-Version`). Now, this message is printed between the assignments `$PossibleVersions = ...` and `$CurrentVersion = ...` in both cases.

I assumed that was an oversight, which is why I added the `Write-Host ...` - or was it intentional?

MichaelKetting: Ah there it is. Sometimes, the diff-viewer isn't very helpful...

ReleaseProcessScript/tools/Core/git_base_functions.ps1
raffaelfoidl: Do I understand you correctly that the comments (as they are) are acceptable? Including the "commit-ish" in the comment text?

I do not like the missing whitespace between comment-hash and comment text either. However, that's how the comments look like across all files, I just stuck to the style. Do you want me to do a global search & replace (`"#X" -> "# X"`), replace it only for the comments in this file or leave it as it is?

MichaelKetting: okay, fine. we already have comments that are stuck to the hash. okay. Perhaps a cleanup task would be appreciated by the maintainers, then...

And yes. text itself is done. Please re-request once all comments are done. (don't want to check this myself)

ReleaseProcessScript/tools/Core/main_helper_functions.ps1
```diff
@@ -69,27 +100,28 @@ function Get-Develop-Current-Version ($StartReleasebranch)
...
 
-function Get-Hotfix-Current-Version ($LastVersion, $StartReleasePhase)
+function Get-Hotfix-Most-Recent-Version ()
 {
-  if ($StartReleasePhase)
+  #Get current support branch from name and version of current branch
+  $CurrentBranchName = Get-Current-Branchname
+  $CurrentBranchFullVersion = Parse-Version-From-BranchName $CurrentBranchName
+  $CurrentBranchMajorMinorVersion = Get-Major-Minor-From-Version $CurrentBranchFullVersion
+  $CurrentSupportBranchName = "support/v$($CurrentBranchMajorMinorVersion)"
```
MichaelKetting, line 113: I think this might be overkill. we already know the version for the hotfix branch. we don't need to get the tags from the support-branch as well.

ReleaseProcessScript/tools/Core/main_helper_functions.ps1
```diff
@@ -49,6 +49,37 @@ function Get-Develop-Current-Version ($StartReleasebranch)
     return Read-Current-Version
   }
 
+  $MostRecentVersion = Get-Develop-Most-Recent-Version
+
```
MichaelKetting, line 53: extra whitespace

MichaelKetting (changes requested):

ReleaseProcessScript/tools/Core/main_helper_functions.ps1
raffaelfoidl: We don't get the tags from the support branch. But I agree, that might be a bit confusing. We need two pieces of information (the other two variables are auxiliary):

1. **Version of support branch**: This acts as fallback version if there is not at least one version tag on the hotfix branch (`else` branch of the `if` in this function).
2. **Name of support branch**: The support branch's name is required as `$To` (i. e. second) parameter in  
 `Get-Last-Version-Of-Branch-From-Tag-Exists "HEAD" $CurrentSupportBranchName`  
`Get-Last-Version-Of-Branch-From-Tag "HEAD" $CurrentSupportBranchName`. If there is a version tag in the "commit range" from HEAD (at this stage pointing to the hotfix branch) until the tip of the support branch, then return the highest version. This `$CurrentSupportBranchname` as second parameter prevents accidentally retrieving a version tag from "below" the fork point of the hotfix branch to release from.

MichaelKetting: Okay, you need the support branch as the lower boundary of the search.

MichaelKetting (approved):

PR #23, opened 2021-02-18
RMSRCBUILD-143 Release Process Script Pushes All Tags Instead of Just the Newly Created Tag

raffaelfoidl:
https://re-motion.atlassian.net/browse/RMSRCBUILD-143

### DIFF
--- a/ReleaseProcessScript/tools/Core/git_base_functions.ps1
+++ b/ReleaseProcessScript/tools/Core/git_base_functions.ps1
@@ -83,7 +83,7 @@ function Is-On-Branch ($Branchname)
 	return $FALSE
 }
 
-function Push-To-Repos ($Branchname, $WithTags)
+function Push-To-Repos ($Branchname, $Tagname)
 {
   $ConfigFile = Get-Config-File
 
@@ -97,9 +97,9 @@ function Push-To-Repos ($Branchname, $WithTags)
   $BeforeBranchname = Get-Current-Branchname
   git checkout $Branchname 2>&1 --quiet
 
-  if ($WithTags)
+  if ( (-not [string]::IsNullOrEmpty($Tagname)) -and (-not (Get-Tag-Exists $Tagname)))
   {
-    $PostFix = "--follow-tags"
+    throw "Tag with name '$($Tagname)' does not exist, abort pushing branch and tag."
   }
 
   $RemoteNames = $ConfigFile.settings.remoteRepositories.remoteName
@@ -141,7 +141,7 @@ function Push-To-Repos ($Branchname, $WithTags)
         }
       } 
 
-      & git push $SetUpstream $RemoteName $Branchname $PostFix 2>&1 | Write-Host
+      & git push $SetUpstream $RemoteName $Branchname $Tagname 2>&1 | Write-Host
     }
   }
     
--- a/ReleaseProcessScript/tools/Core/main_functions.ps1
+++ b/ReleaseProcessScript/tools/Core/main_functions.ps1
@@ -472,7 +472,7 @@ function Continue-Patch-Release ()
     return
   }
 
-  Push-To-Repos $MergeTargetBranchName $TRUE
+  Push-To-Repos $MergeTargetBranchName $Tagname
   Push-To-Repos "release/v$($CurrentVersion)"
 }
 
@@ -551,5 +551,5 @@ function Continue-Pre-Release ()
   }
 
   Push-To-Repos $PrereleaseBranchname
-  Push-To-Repos $CurrentBranchname $TRUE
+  Push-To-Repos $CurrentBranchname $Tagname
 }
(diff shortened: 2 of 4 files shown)


ReleaseProcessScript/tools/Core/git_base_functions.ps1
```diff
@@ -63,9 +63,9 @@ function Push-To-Repos ($Branchname, $WithTags)
   $BeforeBranchname = Get-Current-Branchname
   git checkout $Branchname 2>&1 --quiet
 
-  if ($WithTags)
+  if ( (-not [string]::IsNullOrEmpty($Tagname)) -and (-not (Get-Tag-Exists $Tagname)))
   {
-    $PostFix = "--follow-tags"
+    throw "Tag with name '$($Tagname)' does not exist, abort pushing branch and tag."
```
raffaelfoidl, line 68: There are multiple options here and I do not know which one is preferable. One could throw if a non-existent tagname is given (just like in the code above) or just "log" that the tag does not exist and only the branch will be pushed. In the case of throwing, one could also argue that it would be nice to checkout `$BeforeBranchname` before throwing (btw, should I rename it to `$PreviousBranchname`?).

There possibly are more even alternatives. I suppose it would be better _not_ to throw and just to push no tag - or what should be the way to go?

ReleaseProcessScript/tools/ReleaseProcessScript.psm1
```diff
@@ -112,7 +112,7 @@ function Push-Remote-Repositories ()
     [switch]$WithTags
   )
 
-  Push-To-Repos $Branchname $WithTags
+  Push-To-Repos $Branchname
```
raffaelfoidl, line 115: I thought it would be fitting to push no tags for the time being (until https://re-motion.atlassian.net/browse/RMSRCBUILD-152) such that the entry point still "works" (with the tag(s) being pushed manually) - especially since the $WithTags switch (formerly second parameter of `Push-To-Repos`) does not exist anymore. One would get an error that there is no tag/refspec `True`/`False` when trying to push.

Or should I better leave it untouched?

ReleaseProcessScript/tools/Core/git_base_functions.ps1
Stefan-Ilic: Actually, I would argue that if somebody made the effort to pass a specific branch name (even though in this case no real person is directly calling this function), if that branch name doesn't exist it's most likely some kind of error worthy of an exception

ReleaseProcessScript/tools/ReleaseProcessScript.psm1
Stefan-Ilic: If you remove the parameter, you also need to remove the parameter's help text. But I have an entirely different question: After your changes, is the ReleaseProcessScript in a "consistent state" now? Because to me, it almost looks like it would't be, because we have changed the infrastructure without changing the entry point because we "don't quite know what to do exactly". If anything, we probably should look at [RMSRCBUILD-152](https://re-motion.atlassian.net/browse/RMSRCBUILD-152) first, before merging this PR and actually make a final design decision (or have them both ready so we can merge them in quick succession). What does @MichaelKetting think?

Stefan-Ilic (changes requested):

ReleaseProcessScript/tools/ReleaseProcessScript.psm1
raffaelfoidl: You are right regarding the parameter's help text. However, in the current form, the parameter is still there, but unused (which might be even worse) - that's why I temporarily kept the help text since I was unsure how to proceed with this entrypoint.

I created [RMSRCBUILD-152](https://re-motion.atlassian.net/browse/RMSRCBUILD-152) after a conversation with @MichaelKetting asking what to do with this entrypoint because we cannot know the tag to be pushed without further introspection (e. g. instead of directly calling `Push-To-Repos` retrieving the tag name based on the branch name or the like). He suggested creating the ticket and giving thought to this issue a bit later on.

His main reasoning was that `Push-Remote-Repositories` is an entrypoint that's not used that often and manually pushing a tag afterwards (or simply knowing not to use this entrypoint for the time being) is not a big burden. On the other hand, it might be quite cumbersome if (many) tags that coincidentally are lying around locally are pushed to the remote and one has to delete them all afterwards. That's why, I think, we do want to merge this PR before starting work on [RMSRCBUILD-152](https://re-motion.atlassian.net/browse/RMSRCBUILD-152).

MichaelKetting: I don't remember if I wanted to drop the tags at this unrelated function. It may have been dicussed but there's not need tp change the behavior of unrelated code in this PR. The Push-to-repos is not used anyway. So long as that's the case, we can fix / drop Push-to-repos separately and keep things clean.

raffaelfoidl: I have reverted the changes to this file in order to concentrate on this issue later on separately.

Stefan-Ilic (approved):

MichaelKetting (approved):

PR #24, opened 2021-03-11
RMSRCBUILD-155: Release Process Script Does Not Handle Errors When Releasing Jira Version

raffaelfoidl:
https://re-motion.atlassian.net/browse/RMSRCBUILD-155

### DIFF
--- a/ReleaseProcessScript/tools/Core/jira_functions.ps1
+++ b/ReleaseProcessScript/tools/Core/jira_functions.ps1
@@ -64,6 +64,7 @@ function Jira-Create-Version ($Version)
   catch 
   {
     [System.InvalidOperationException]
+    # Rethrow C# exception because the PS script does not terminate otherwise
     throw $_.Exception.Message
   }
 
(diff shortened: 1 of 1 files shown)


Stefan-Ilic (approved):

ReleaseProcessScript/tools/Core/jira_functions.ps1
```diff
@@ -84,7 +84,14 @@ function Jira-Get-Current-Version ()
...
   Add-Authentication $JiraGetVersion
 
-  $JiraGetVersion.Execute()
+  try
+  {
+    $JiraGetVersion.Execute()
+  }
+  catch
+  {
+    throw $_.Exception.Message
```
MichaelKetting, line 93: I.C.

Probably should add a code-comment for these two blocks because it's really not that obvious without knowning the history.

MichaelKetting (changes requested):

ReleaseProcessScript/tools/Core/jira_functions.ps1
raffaelfoidl: Understandable. However, I did not do so because these `try-catch` blocks have been in every function from `jira_functions.ps1` (except for the two methods affected by the changes in this PR). Do you want me to add a comment to each `try-catch` block or e. g. a comment at the top of the file that addresses every jira function?

MichaelKetting (approved):

PR #30, opened 2021-06-11
RMSRCBUILD-162: Calculation of next release version does not work for hotfix branches

raffaelfoidl:
https://re-motion.atlassian.net/browse/RMSRCBUILD-162

### DIFF
--- a/ReleaseProcessScript/tools/Core/main_helper_functions.ps1
+++ b/ReleaseProcessScript/tools/Core/main_helper_functions.ps1
@@ -68,9 +68,29 @@ function Get-Hotfix-Current-Version ($StartReleasePhase)
   }
   else
   {
+    $HotfixBranchName = Get-Current-Branchname
     $MostRecentVersion = Get-Hotfix-Most-Recent-Version
 
-    $PossibleVersions = Get-Possible-Versions-Hotfix $MostRecentVersion $true
+    #Get current support branch from name and version of current branch
+    $CurrentBranchFullVersion = Parse-Version-From-BranchName $HotfixBranchName
+    $CurrentBranchMajorMinorVersion = Get-Major-Minor-From-Version $CurrentBranchFullVersion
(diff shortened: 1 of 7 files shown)


ReleaseProcessScript/tools/Core/main_helper_functions.ps1
```diff
@@ -68,9 +68,29 @@ function Get-Hotfix-Current-Version ($StartReleasePhase)
...
+    # We want the following behaviour for the MeantForCurrentVersion flag of Get-Possible-Verions-Hotfix:
+    #
+    #   PreReleaseExists | NoReleaseExists |   MeantForCurrentVersion
+    #  ------------------|-----------------|-----------------------------
+    #   False            | False           | False
+    #   False            | True            | True
+    #   True             | False           | True
+    #   True             | True            | Impossible (fallback False)
+    #
+    # => this is equivalent to PreReleaseExists XOR NoReleaseExists
```
MichaelKetting, line 88: I think I like it. And despite the double-negation for NoReleaseExists, the not-negated XOR is indeed a nice simplification.

MichaelKetting: Regardless of the mathematical beauty, I do wonder if the Impossible case couldn't also do a fallback to True and thus make this into a simple OR setup. But that shouldn't hold up the merge.

MichaelKetting (approved):

iam-mholle (approved):
Nice job 👍

MichaelKetting (commented):
After testing in production, the soltution doesn't work for 0-patches.

commit 1:
support/v2.28
support/v2.29
tags/v2.28.0

commit 2: 
hotfix/v2.29.0

script only offers version for v2.28 instead of going with v2.29.

Updating the local repo by removing the local "support/v2.28" does not help with "hotfix/v2.29.0".

PR #34, opened 2022-02-21
RMSRCBUILD-191 CompileTestProjects Target

flo003:


BuildScript.Nuke/Build.Compile.cs
```diff
@@ -7,37 +7,74 @@
...
+          );
+        });
+  }
+
+  private void RestoreProject (ProjectMetadata projectFile)
+  {
+    MSBuild(
+              s => s
+                  .SetTargetPath(projectFile.Path)
+                  .SetTargets("Restore")
```
iam-mholle, line 69: Formatting is off

BuildScript.Nuke/Build.Compile.cs
```diff
@@ -7,37 +7,74 @@
...
+    );
+  }
+
+  private string[] GetCompileTargets (ProjectMetadata project)
+  {
+    var targets = new List<string> { "Build" };
+    if (project.IsMultiTargetFramework)
+      targets.Add("DispatchToInnerBuilds");
+    return targets.ToArray();
```
iam-mholle, line 78: Wouldn't this be either/or?

iam-mholle (changes requested):

BuildScript.Nuke/Build.Compile.cs
Gigi-White: In MSBuild Log both are used

PR #49, opened 2022-03-14
feature/RMSRCBUILD-217 add license header to nuke files

jxhxnnx:


iam-mholle (changes requested):
Please also add the license header to the dotsettings, for non-licenseheadermanager users :)

iam-mholle (approved):

PR #55, opened 2022-03-29
RMSRCBUILD-245 Optimize build order

Gigi-White:


BuildScript.Nuke/BaseBuild.Compile.cs
```diff
@@ -30,13 +34,23 @@ namespace Remotion.BuildScript;
 [SuppressMessage("ReSharper", "MemberCanBePrivate.Global")]
 public partial class BaseBuild : NukeBuild
 {
+  private Task<IReadOnlyCollection<Project>>? _sortedProjectList;
+
+  private Task<IReadOnlyCollection<Project>> SortedProjectList
+  {
+    get { return _sortedProjectList ??= GetSortedProjects(Solution.Path); }
+  }
```
iam-mholle, line 42:
```suggestion
  private IReadOnlyCollection<Project>? _sortedProjectList;
  private IReadOnlyCollection<Project> SortedProjectList => _sortedProjectList ??= GetSortedProjects(Solution.Path).GetAwaiter().GetResult();
```

This should be safe to do here.

BuildScript.Nuke/BaseBuild.Compile.cs
```diff
@@ -56,59 +70,90 @@ public partial class BaseBuild : NukeBuild
...
-                .SetInformationalVersion(SemanticVersion.GetAssemblyInformationalVersion(projectFile.Configuration, AdditionalBuildMetadata))
-        )
-    );
+    var projectFilesDebugTargets = projectFiles
+        .Where(file => file.Configuration == "Debug").ToList();
+    var projectFilesReleaseTargets = projectFiles
+        .Where(file => file.Configuration == "Release").ToList();
+    if (projectFilesDebugTargets.Count > 0) await CompileProject(projectFilesReleaseTargets, "Debug", "Current");
+    
+    if (projectFilesReleaseTargets.Count > 0) await CompileProject(projectFilesReleaseTargets, "Release", "Current");
```
iam-mholle, line 97: The tools version from the other PR should be used here I guess. (Instead of `Current`). Just a reminder to adapt the second PR in case this gets merged first.

BuildScript.Nuke/BaseBuild.Compile.cs
```diff
@@ -56,59 +70,90 @@ public partial class BaseBuild : NukeBuild
...
+
+    filteredProjects.ForEach(project =>
+    {
+      MSBuild(s => s
+          .SetProperty(MSBuildProperties.RestorePackagesConfig, true)
+          .SetProperty(MSBuildProperties.SolutionDir, Directories.Solution)
+          .SetTargetPath(project.FilePath)
+          .SetTargets(MSBuildTargets.Restore)
+      );
+    });
```
iam-mholle, line 157: Isn't the restore order irrelevant? Why are the sorted projects needed here?

iam-mholle (changes requested):

BuildScript.Nuke/BaseBuild.Compile.cs
iam-mholle: The other one got merged first, so please adapt :)

BuildScript.Nuke/BaseBuild.Compile.cs
```diff
@@ -66,49 +73,63 @@ public partial class BaseBuild : NukeBuild
...
   }
 
   private string GetToolPath ()
   {
-    var toolPath = MSBuildTasks.MSBuildPath;
+    var toolPath = MSBuildPath;
     var editions = new[] { "Enterprise", "Professional", "Community", "BuildTools", "Preview" };
     if (!string.IsNullOrEmpty(MsBuildPath))
       toolPath = MsBuildPath;
```
iam-mholle, line 135: you could just return the `MsBuildPath` here, or return the result of the statement in the `else if` block. Anyway, what if no property is set? Then the returned `toolPath` would be null. I think it would be OK to `Assert.Fail()` if nothing is available. Additionally, the code should also work if only VS2022 Professional is installed, but in this case, the installation would not be found because the non-X86 folder is not checked. Please create a follow up issue for this.

iam-mholle (approved):

PR #58, opened 2022-03-31
RMSRCBUILD-264 Fix version for local builds

Gigi-White:


BuildScript.Nuke/SemanticVersion.cs
```diff
@@ -72,9 +72,12 @@ public class SemanticVersion
   /// </remarks>
   public string VersionWithoutPreRelease { get; }
 
-  public SemanticVersion (string version)
+  public SemanticVersion (string version, bool isLocalBuild)
   {
     Version = version;
+    if (isLocalBuild)
+      Version = $"{Version.Split('-')[0]}-x.9.{DateTime.Now:yyMMdd-HHmmss}";
```
iam-mholle, line 79: `versionMatch` contains all parts needed to build the local version, please don't split here.

BuildScript.Nuke.UnitTests/SemanticVersionTest.cs
```diff
@@ -109,6 +111,71 @@ public void DependDBProjectVersion_WithValidVersion_ReturnsDepenDBProjectVersion
...
+    DateTime parseDateTime;
+
+    var semanticVersion = new SemanticVersion(version, true);
+    var assemblyInformationVersion = semanticVersion.GetAssemblyInformationalVersion(configurationId, additionalBuildMetadata);
+
+    var splittedVersion = assemblyInformationVersion.Split('.');
+    var actualVersionDateTime = splittedVersion.Single(x => x.Contains('+')).Split('+').First();
+    var isValidDaTime = DateTime.TryParseExact(actualVersionDateTime, format, new CultureInfo("en-US"), DateTimeStyles.None, out parseDateTime);
+
+    Assert.That(assemblyInformationVersion, Does.Contain("-x.9."));
```
iam-mholle, line 158: I think it would be better to use regex expressions to assert the version structure, what do you think?

iam-mholle, line 158: I think it would be better to use regex expressions to assert the version structure, what do you think?

iam-mholle (changes requested):

iam-mholle (approved):

PR #66, opened 2022-05-16
RMSRCBUILD-268 Add jira integration

Frazerus:
https://re-motion.atlassian.net/browse/RMSRCBUILD-268


ReleaseProcessAutomation/Jira/JiraCredentialManager.cs
```diff
@@ -0,0 +1,138 @@
...
+    }
+    catch
+    {
+      // ignored
+    }
+
+    CredentialManager.SaveCredentials(target, cred);
+  }
+
+  private bool CheckJiraAuthentication (Credentials credentials)
```
a-ctor, line 120: With the bool setup as return value we are losing the detailed exception info created by `CheckJiraCredentials`. Maybe we could rewrite this to throw our new `JiraAuthenticationException` with an inner exception? What do you think?

a-ctor, line 120: With the bool setup as return value we are losing the detailed exception info created by `CheckJiraCredentials`. Maybe we could rewrite this to throw our new `JiraAuthenticationException` with an inner exception? What do you think?

ReleaseProcessAutomation/Jira/JiraCredentialManager.cs
```diff
@@ -0,0 +1,138 @@
...
+using ReleaseProcessAutomation.ReadInput;
+using Serilog;
+using Spectre.Console;
+
+namespace ReleaseProcessAutomation.Jira;
+
+public interface IJiraCredentialManager
+{
+  Credentials GetCredential (string target);
+}
```
a-ctor, line 31: I know that the other files with the steps always included the interface together in the same file. Usually we have them in separate file so for new code we should do that. I assume it was done with the steps to prevent having twice the files in the folder :D

a-ctor, line 31: I know that the other files with the steps always included the interface together in the same file. Usually we have them in separate file so for new code we should do that. I assume it was done with the steps to prevent having twice the files in the folder :D

ReleaseProcessAutomation/Jira/JiraCredentialManager.cs
```diff
@@ -0,0 +1,138 @@
...
+
+public class JiraCredentialManager
+    : IJiraCredentialManager
+{
+  private IJiraEntrancePoint _jira;
```
a-ctor, line 36:
```suggestion
  private IJiraEntrancePoint _jira;
```
```suggestion
  private readonly IJiraEntrancePoint _jira;
```

a-ctor, line 36:
```suggestion
  private readonly IJiraEntrancePoint _jira;
```

ReleaseProcessAutomation.Tests/Jira/JiraCredentialManagerTests.cs
```diff
@@ -0,0 +1,165 @@
+using System.Net;
+using AdysTech.CredentialManager;
```
a-ctor, line 2: I think we are also missing the unit test from the previous project?

ReleaseProcessAutomation/Jira/JiraGetEarliestUnreleasedVersion.cs
```diff
@@ -0,0 +1,54 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using JetBrains.Annotations;
+using ReleaseProcessAutomation.Jira.ServiceFacadeImplementations;
+using ReleaseProcessAutomation.Jira.ServiceFacadeInterfaces;
+
+namespace ReleaseProcessAutomation.Jira
+{
+  public class JiraGetEarliestUnreleasedVersion : JiraTask
```
a-ctor, line 27: Since this class is not used I think we can just delete the dead code. Same goes for other parts which are not referenced. If somebody needs it in the future they can always use the history to find it. But I don't think it adds anything otherwise.

Or is there a future task where they are needed again?

cc @MichaelKetting

ReleaseProcessAutomation/Jira/JiraGetLatestReleasedVersion.cs
```diff
@@ -0,0 +1,51 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using JetBrains.Annotations;
+using ReleaseProcessAutomation.Jira.ServiceFacadeImplementations;
+using ReleaseProcessAutomation.Jira.ServiceFacadeInterfaces;
+
+namespace ReleaseProcessAutomation.Jira
+{
+  public class JiraGetLatestReleasedVersion : JiraTask
```
a-ctor, line 27: ded

ReleaseProcessAutomation/Jira/JiraVersionCreator.cs
```diff
@@ -0,0 +1,112 @@
...
+
+namespace ReleaseProcessAutomation.Jira;
+
+public class JiraVersionCreator : JiraTask
+{
+
+  public JiraVersionCreator ([CanBeNull] string? jiraUsername, [CanBeNull] string? jiraPassword) : 
+      base(jiraUsername, jiraPassword) { }
+
+  public void CreateNewVersion (string jiraUrl, string jiraProject, string versionPattern, bool sortVersion, int versionComponentToIncrement, DayOfWeek versionReleaseWeekday)
```
a-ctor, line 35: ded?

ReleaseProcessAutomation/Jira/JiraVersionReleaser.cs
```diff
@@ -0,0 +1,58 @@
...
+    IJiraProjectVersionService service = new JiraProjectVersionService(restClient);
+    service.ReleaseVersionAndSquashUnreleased(versionID, nextVersionID, jiraProjectKey);
+  }
+  
+  
+  
+  
+
+
+}
```
a-ctor, line 58:
```suggestion
  }
}
```

ReleaseProcessAutomation/Jira/JiraTask.cs
```diff
@@ -0,0 +1,39 @@
...
+using System;
+using RestSharp.Authenticators;
+
+namespace ReleaseProcessAutomation.Jira
+{
+  public abstract class JiraTask
+  {
+    protected IAuthenticator Authenticator;
+
+    protected JiraTask (string? jiraUsername, string? jiraPassword)
```
a-ctor, line 27: The more I look at `JiraTask` the more I want to remove it completely :D 

Its only use case here is the creation of the authenticator, which seems wrong to begin with. The only usages for the authenticator are the creation of the `JiraRestClient`. So we are probably better off if we remove this class and feed in the `JiraRestClient` directly in the subclasses.

If we want to go next level we can also go one step further as the `JiraRestClient` is only used to create other services. We could move that logic into an interface, for example, `IJira` and have it create the services for us. This way the jira task no longer need to care about auth/rest whatever and simply create the services they need to do their jobs

ReleaseProcessAutomation/Jira/ServiceFacadeImplementations/IJiraRestApi.cs
```diff
@@ -0,0 +1,25 @@
...
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the 
+// License for the specific language governing permissions and limitations
+// under the License.
+//
+
+using System;
+
+namespace ReleaseProcessAutomation.Jira.ServiceFacadeImplementations
+{
+  interface IJiraRestApi
```
a-ctor, line 22: ded?

ReleaseProcessAutomation/Jira/ServiceFacadeImplementations/JiraIssue.cs
```diff
@@ -0,0 +1,59 @@
...
+// License for the specific language governing permissions and limitations
+// under the License.
+//
+
+using System;
+using System.Collections.Generic;
+
+namespace ReleaseProcessAutomation.Jira.ServiceFacadeImplementations
+{
+  public class JiraIssue
```
a-ctor, line 23: I don't know how much benefit nullability info for this DTOs provide. Because whenever we use the (see JiraIssueService@42) we need to check for null all the time. Maybe it is better to remove it here and just go with the REST Api should return correct results. Or check them once when receving them.

a-ctor (changes requested):

ReleaseProcessAutomation/Jira/JiraGetEarliestUnreleasedVersion.cs
MichaelKetting: no, unused stuff can be dropped.

ReleaseProcessAutomation/Jira/JiraEntrancePoint.cs
```diff
@@ -0,0 +1,140 @@
...
+namespace ReleaseProcessAutomation.Jira;
+
+public interface IJiraEntrancePoint
+{
+  void CreateAndReleaseJiraVersion (SemanticVersion currentVersion, SemanticVersion nextVersion, bool squashUnreleased = false);
+
+  void CheckJiraCredentials (Credentials credentials);
+}
+
+public class JiraEntrancePoint : IJiraEntrancePoint
```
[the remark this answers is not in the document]
Frazerus: I'm not sure that `CreateAndReleaseJiraVersion` should be made into a step as it more closely resembles the behaviour of `MsBuild`, which is not a step aswell. 

Making `CreateAndReleaseJiraVersion` a step would also create several problems it can be called in different parts of the application, therefore not always having a following step, as this would completely break up the order of how  things are done. 

I would suggest renaming the `JiraEntrancePoint` to `CreateAndReleaseJiraVersion` like you mentioned and renaming it's only public method to `Execute`.

ReleaseProcessAutomation/Jira/JiraTask.cs
Frazerus: The `JiraTask `is gone now, but I'm not quite sure what you meant in the last paragraph. Could you add a more detailed description of what you mean?

ReleaseProcessAutomation.Tests/Jira/JiraCredentialManagerTests.cs
Frazerus: I wouldn't know where they are as I think I already included all of them? I think the test s are just missing in general. I'm creating unit tests for `JiraEntrancePoint` and possibly `JiraCredentialManager`

ReleaseProcessAutomation/Jira/ServiceFacadeImplementations/JiraIssue.cs
Frazerus: Quite understandable as this was simply a change to enable me to work on different aspects. I fully agree on removing nullability as it brings no clear advantages IMO. 

I do, however, not know the best way of removing nullability in this case. Leave them as properties and initialize them? Create constructors? What would be the proper way?

ReleaseProcessAutomation/Jira/JiraEntrancePoint.cs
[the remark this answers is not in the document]
a-ctor: Yes looking at it again making it a step is probably not the best idea :D

We should split these two methods into two classes maybe `IJiraFunctionality` and `IJiraAuthenticator` or something. Since we are not converting the class to a step it might get new functionality in the future, which is why I would prefer to stay away from `CreateAndReleaseJiraVersion`.

Since we have some shared logic like `JiraUrlWithPostfix` we probably want to move that into another class like `JiraMetadata` and have the other ones depend on it for the functionality.

ReleaseProcessAutomation/Jira/JiraTask.cs
a-ctor: If you look at the usages of the `JiraRestClient` it is always passed into the other services:

    var jiraProjectVersionFinder = new JiraProjectVersionFinder (jiraRestClient);

This way, we always have hard dependencies on them and it is more difficult to test. What we would do is create an interfaces with implementation (the `IJira`) which contains all the services that are used for the interaction with Jira.

Instead of the code above the method would receive a `IJira` instead of the `JiraRestClient` and then access the `JiraProjectVersionFinder` using `jira.JiraProjectVersionFinder`. This way, we can use the interfaces and do not need to depend on the actual classes making testing easier while also removing the need for the method to know about a `JiraRestClient`.

ReleaseProcessAutomation/Jira/ServiceFacadeImplementations/JiraIssue.cs
a-ctor: Probably the best way is to add a `#nullable disable` directive at the top of the files that contain DTOs. We could also do `= default!;` initialization but that is more verbose.

ReleaseProcessAutomation.Tests/Jira/JiraCredentialManagerTests.cs
a-ctor: I just missed them I think, my bad :D

ReleaseProcessAutomation/Jira/JiraTask.cs
Frazerus: There is now an `IJira` interface with properties to be used like `_jira.VersionCreator.CreateVersion()`. The properties are however not inserted into the implementation, they are created at runtime. This is because otherwise we would start the application and would immediately be asked for the JiraCredentials. 

To insert them we could either ignore that problem, or create some sort of `Initialize` Method in all of the services. Do we want that?

a-ctor: Constructors should not contain much more logic than setting their fields. If we prompt the user in the constructor we should change that. Looking at the code I think it is best to introduce a provider for the `JiraRestClient`.

The provider interface contains a single `JiraRestClient GetJiraRestClient()` method and the implementation does the prompting, creation, and probably also caches the instance for future requests. The `IJira` implementation then receives the provides and stores it and uses it in its methods to retrieve an instance of the `JiraRestClient`.

This way, we only set a field in the constructor and the prompting is done the first time an actual action is done.

Build/Customizations/Projects.props
```diff
@@ -9,11 +9,11 @@
   </PropertyGroup>
   <ItemGroup>
     <ReleaseProjectFiles Include="$(SolutionDirectory)\BuildScript\Buildscript.csproj" />
-    <ReleaseProjectFiles Include="$(SolutionDirectory)\ReleaseProcessScript\ReleaseProcessScript.csproj" />
+    <ReleaseProjectFiles Include="$(SolutionDirectory)\ReleaseProcsessAutomation\ReleaseProcessAutomation.csproj" />
     <UnitTestProjectFiles Include="$(SolutionDirectory)\BuildScript.UnitTests\BuildScript.UnitTests.csproj">
       <TestConfiguration>$(NormalTestConfiguration)</TestConfiguration>
     </UnitTestProjectFiles>
-    <UnitTestProjectFiles Include="$(SolutionDirectory)\ReleaseProcessScript.UnitTests\ReleaseProcessScript.UnitTests.csproj">
+    <UnitTestProjectFiles Include="$(SolutionDirectory)\ReleaseProcessAutomation.UnitTests\ReleaseProcessAutomation.UnitTests.csproj">
```
a-ctor, line 16: If we replace the solutions here we should also remove the projects. Is there another task to do that? Otherwise, I think we can also remove the projects. We probably want to have that in a separate commit in the end (so one adds the new projects, and one removes the old projects and deletes them)

Build/Customizations/releaseProcessScript.config
```diff
@@ -2,7 +2,7 @@
 <settings>
   <jira>
     <jiraUrl>https://re-motion.atlassian.net</jiraUrl>
-    <jiraProjectKey>RMSRCBUILD</jiraProjectKey>
+    <jiraProjectKey>DEFINETLYNOTAPROJECTKEY</jiraProjectKey>
```
a-ctor, line 5: Now that we made sure that the tests do not use the actual jira instance but mocks is there any reason to keep this change?

ReleaseProcessAutomation.IntegrationTests/GitBackedTests.cs
```diff
@@ -20,18 +20,75 @@
...
+  }
+
+  protected void AddRandomFile (string directoryPath)
+  {
+    var random = new Random().Next();
+
+    var path = Path.Join(directoryPath, random.ToString());
+
+    using var fs = File.Create(path);
+    fs.Dispose();
```
a-ctor, line 65: Either `using` or `.Dispose()` is needed and I think that `.Dispose()` is clearer.

```suggestion
    var fs = File.Create(path);
    fs.Dispose();
```

ReleaseProcessAutomation.IntegrationTests/ReleaseProcessScript.Test.Config
```diff
@@ -2,7 +2,7 @@
 <settings>
     <jira>
         <jiraUrl>https://re-motion.atlassian.net</jiraUrl>
-        <jiraProjectKey>GHTEST</jiraProjectKey>
+        <jiraProjectKey>DEFINETLYNOTAPROJECTKEY</jiraProjectKey>
```
a-ctor, line 5: Same questions as above, do we still need that?

ReleaseProcessAutomation.IntegrationTests/ReleaseProcessScript.Test.Config
```diff
@@ -24,7 +24,8 @@
         <fileName>Build/Customizations/Version.props</fileName>
     </developStableMergeIgnoreList>
     <msBuildSettings>
-        <msBuildPath>C:\Program Files\Microsoft Visual Studio\2022\Professional\MSBuild\Current\Bin\MSBuild.exe</msBuildPath>
+        <msBuildPath>C:\Program Files\Microsoft Visual Studio\2022\Professional\MSBuild\Current\Bin\MSBuild.exe
+        </msBuildPath>
```
a-ctor, line 28: We should revert that

ReleaseProcessAutomation.UnitTests/Jira/JiraCredentialManagerTests.cs
```diff
@@ -0,0 +1,213 @@
...
+using ReleaseProcessAutomation.ReadInput;
+using Spectre.Console.Testing;
+
+namespace ReleaseProcessAutomation.UnitTests.Jira;
+
+[TestFixture]
+public class JiraCredentialManagerTests
+{
+  [SetUp]
+  public void Setup ()
```
a-ctor, line 19: We should reorder these statements. Here are the general rules we try to adhere to:

- Constants > Static stuff > Instance stuff.
  - stuff is always: fields (including properties that act as state [auto properties]) > constructor > properties > methods
- public before private e.g. private instance fields then public instance fields
- Special NUnit methods like the setup and teardown usually come before any of the tests

This makes it easier to quickly find what you are looking for.

ReleaseProcessAutomation.UnitTests/ReleaseProcessAutomation.UnitTests.csproj
```diff
@@ -36,4 +34,8 @@
         </None>
     </ItemGroup>
 
+    <ItemGroup>
+      <Folder Include="Git" />
+    </ItemGroup>
+
```
a-ctor, line 40:
```suggestion
```

ReleaseProcessAutomation.UnitTests/Steps/InitialBranching/BranchFromDevelopStepTests.cs
```diff
@@ -0,0 +1,255 @@
...
+namespace ReleaseProcessAutomation.UnitTests.Steps.InitialBranching;
+
+[TestFixture]
+internal class BranchFromDevelopStepTests
+{
+  [SetUp]
+  public void Setup ()
+  {
+    ;
+
```
a-ctor, line 36:
```suggestion
```

ReleaseProcessAutomation.UnitTests/Steps/InitialBranching/BranchFromDevelopStepTests.cs
```diff
@@ -0,0 +1,255 @@
...
+    _releaseAlphaBetaMock = new Mock<IReleaseAlphaBetaStep>();
+    _releaseAlphaBetaMock.Setup(_ => _.Execute(It.IsAny<SemanticVersion>(), It.IsAny<string>(), false, false)).Verifiable();
+
+    _releaseOnMasterMock = new Mock<IReleaseOnMasterStep>();
+    _releaseOnMasterMock.Setup(_ => _.Execute(It.IsAny<SemanticVersion>(), "", false, false, false)).Verifiable();
+  }
+
+  private Mock<IReleaseAlphaBetaStep> _releaseAlphaBetaMock;
+  private Mock<IReleaseOnMasterStep> _releaseOnMasterMock;
```
a-ctor, line 45: Move up. Also check the other step tests for the same thing please.

ReleaseProcessAutomation/Jira/Authentication/IJiraAuthenticator.cs
```diff
@@ -0,0 +1,9 @@
+using System;
+using ReleaseProcessAutomation.Jira.CredentialManagement;
+
+namespace ReleaseProcessAutomation.Jira.Authentication;
+
+public interface IJiraAuthenticator
+{
+  public void CheckAuthentication (Credentials credential, string projectKey, string jiraURL);
```
a-ctor, line 8:
```suggestion
  void CheckAuthentication (Credentials credential, string projectKey, string jiraURL);
```

ReleaseProcessAutomation/Jira/Authentication/JiraAuthenticator.cs
```diff
@@ -0,0 +1,23 @@
...
+using RestSharp.Authenticators;
+
+namespace ReleaseProcessAutomation.Jira.Authentication;
+
+public class JiraAuthenticator
+    : IJiraAuthenticator
+{
+  public void CheckAuthentication (Credentials credentials, string projectKey, string jiraURL)
+  {
+    //var jiraRestClient = new JiraRestClient(jiraURL, new HttpBasicAuthenticator(credentials.Username, credentials.Password));
```
a-ctor, line 16:
```suggestion
```

ReleaseProcessAutomation/Jira/CredentialManagement/JiraCredential.cs
```diff
@@ -0,0 +1,33 @@
...
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the 
+// License for the specific language governing permissions and limitations
+// under the License.
+//
+
+using System;
+
+namespace ReleaseProcessAutomation.Jira.CredentialManagement;
+
+public struct Credentials
```
a-ctor, line 22: We should rename the file or the class so that they both match up. I prefer `Credential` and renaming the file

ReleaseProcessAutomation/Jira/CredentialManagement/JiraCredentialManager.cs
```diff
@@ -0,0 +1,143 @@
...
+    var cred = CredentialManager.GetCredentials(target);
+
+    if (cred == null)
+      return AskForCredentials(target);
+
+    var credentials = new Credentials
+                      {
+                          Username = cred.UserName,
+                          Password = cred.Password
+                      };
```
a-ctor, line 62:
```suggestion
    var credentials = new Credentials(cred.UserName, cred.Password);
```

ReleaseProcessAutomation/Jira/CredentialManagement/JiraCredentialManager.cs
```diff
@@ -0,0 +1,143 @@
...
+
+  private Credentials AskForCredentials (string target)
+  {
+    while (true)
+    {
+      var tmpCredentials = new Credentials
+                           {
+                               Username = _inputReader.ReadString("Please enter your Jira username"),
+                               Password = _inputReader.ReadHiddenString("Please enter your Jira password")
+                           };
```
a-ctor, line 85:
```suggestion
     var username = _inputReader.ReadString("Please enter your Jira username"),
     var password = _inputReader.ReadHiddenString("Please enter your Jira password")
      var tmpCredentials = new Credentials(username, password);
```

ReleaseProcessAutomation/Jira/CredentialManagement/JiraCredentialManager.cs
```diff
@@ -0,0 +1,143 @@
...
+  private void SaveCredentials (Credentials tmpCredentials, string target)
+  {
+    var cred = new NetworkCredential(tmpCredentials.Username, tmpCredentials.Password);
+    try
+    {
+      CredentialManager.RemoveCredentials(target);
+    }
+    catch
+    {
+      // ignored
```
a-ctor, line 122: We should document why this is not a problem and why we ignore it. As it stands the comment does not add any value

ReleaseProcessAutomation/ReleaseProcessAutomation.csproj
```diff
@@ -23,14 +23,21 @@
...
         <PackageReference Include="Serilog.Sinks.Debug" Version="2.0.0" />
         <PackageReference Include="Serilog.Sinks.File" Version="5.0.0" />
         <PackageReference Include="Spectre.Console" Version="0.43.0" />
     </ItemGroup>
 
+    <ItemGroup>
+      <Compile Remove="Jira\Utility\IJiraRestClientProviderFactory.cs" />
+      <Compile Remove="Jira\Utility\JiraRestClientProviderFactory.cs" />
+    </ItemGroup>
+
```
a-ctor, line 41:
```suggestion
```

a-ctor (changes requested):
We should also do a test run to make sure that everything works as expected

Build/Customizations/releaseProcessScript.config
Frazerus: I think the better question would be whether we should undo the change as it gives safety against jira dying if someone creates a new test and somehow forgets to check the config that Jira doesnt die. 

So unless we want RM in our test config, I think it would benefit us to have the extra safety.

ReleaseProcessAutomation.IntegrationTests/ReleaseProcessScript.Test.Config
Frazerus: *makes more sense to keep the change

a-ctor: If we do not have any tests that make use of it we can keep the incorrect project key, otherwise we should use a correct one to allow easier testing. You do need to authenticate before you can do anything bad so I don't think it is that bad to have a correct key in here. The `GHTEST` project is also only used for testing so there shouldn't be anything bad possible.

Build/Customizations/Projects.props
Frazerus: Created a task for it.

MichaelKetting:
@Frazerus , I'll wait with my review until the review by @a-ctor is integrated.

PR #76, opened 2022-06-22
Feature/RMSRCBUILD-282 release to support creates new branches

Frazerus:


ReleaseProcessAutomation/Steps/PipelineSteps/ContinueReleasePatchStep.cs
```diff
@@ -92,9 +92,28 @@ public void Execute (SemanticVersion nextVersion, bool noPush, bool onMaster)
...
+    
     if (noPush)
       return;
 
     _pushPatchReleaseStep.Execute(mergeTargetBranchName, tagName, toMergeBranchName);
   }
+
+  private void CreateNewSupportBranch (SemanticVersion nextVersion)
+  {
+    Console.WriteLine("Do you wish to create a new support branch?");
```
MichaelKetting, line 110: Console-Outout would also be a good abstraction to test that the correct messages are written and to prevent console spam during tests.

ReleaseProcessAutomation.Tests/IntegrationTests/ReleaseFromHotfixTests.cs
```diff
@@ -83,13 +83,13 @@ public void ReleaseAlphaBeta_FromHotfix_WithAdditionalCommitAndContinue ()
     Assert.That(act2, Is.EqualTo(0));
 
   }
-
+
```
MichaelKetting, line 86:
```suggestion

```

ReleaseProcessAutomation.Tests/IntegrationTests/ReleaseFromHotfixTests.cs
```diff
@@ -106,9 +106,45 @@ public void ReleaseNewPatch_FromHotfix_ToSupport ()
...
+    //Get next release version from user for jira
+    TestConsole.Input.PushTextWithEnter("1.1.2");
+    //Confirms to create new support branch
+    TestConsole.Input.PushTextWithEnter("y");
+
+    var act = Program.Main(new[] { "Release-Version" });
+    
     AssertValidLogs(correctLogs);
     Assert.That(act, Is.EqualTo(0));
   }
```
MichaelKetting, line 150: This scenario seems weird:
why is there both a support v1.2 and a support v1.1 (including hotfixes) in the expectation?

ReleaseProcessAutomation.Tests/IntegrationTests/ReleaseFromReleaseTests.cs
```diff
@@ -56,13 +56,74 @@ public void ReleaseWithRC_FromSupportHotfixRelease_ReleasesToSupport ()
...
+    TestConsole.Input.PushTextWithEnter("n");
+    
     var act = Program.Main(new[] { "Release-Version" });
 
     AssertValidLogs(correctLogs);
     Assert.That(act, Is.EqualTo(0));
   }
   
+  [Test]
+  public void ReleaseWithRC_FromSupportHotfixReleaseWithNewSupportBranch_ReleasesToSupportAndCreatesSupportBranch ()
```
MichaelKetting, line 69: See other comment.

ReleaseProcessAutomation/Steps/PipelineSteps/ContinueReleasePatchStep.cs
```diff
@@ -92,9 +92,28 @@ public void Execute (SemanticVersion nextVersion, bool noPush, bool onMaster)
 
     GitClient.Checkout(mergeTargetBranchName);
 
+    if (!onMaster)
+    {
+      CreateNewSupportBranch(nextVersion);
```
MichaelKetting, line 97: Why do we not do this on master, too?

MichaelKetting (changes requested):
I think we should do  live demonstration of this PR to answer the open support-branch questions.

ReleaseProcessAutomation.Tests/IntegrationTests/ReleaseFromHotfixTests.cs
Frazerus: As far as I understood the docs, when creating a new support branch at the release of a hotfix, we want the old branch to continue running with a new hotfix version. 

Let's assume we are on hotfix/v1.1.1 and we release this to support/v1.1, we also want to create a new support branch from that release. 
That means we have a new support/v1.2 with a new hotfix/v1.2.0 aswell as the old support/v1.1 with a new hotfix/v1.1.2

I'm not sure this is the behaviour that we want, but I think it's the behaviour outlined in the pptx.

ReleaseProcessAutomation/Steps/PipelineSteps/ContinueReleasePatchStep.cs
Frazerus: Because the only branch releasing to master should be ones coming from the develop branch, where we do not require support branches

Frazerus: Because I forgot that when we release from develop we might need a new support branch on the other side ;)

Frazerus: Except maybe in this case we do not want this behaviour, as releasing a patch to master is supported but not intended.  We do however require this functionality in other steps releasing to master.

MichaelKetting: Better create one and delete than needing to do all this manually (optimize for the frequent case)

ReleaseProcessAutomation.Tests/IntegrationTests/ReleaseFromHotfixTests.cs
MichaelKetting: > That means we have a new support/v1.2 with a new hotfix/v1.2.0 aswell as the old support/v1.1 with a new hotfix/v1.1.2
This interpretation is incorrect. The support/v1.2 can be created as a fork from support v1.1, but only on demand, not automatically.

ReleaseProcessAutomation/Steps/PipelineSteps/ContinueReleasePatchStep.cs
Frazerus: Ok, will do!

ReleaseProcessAutomation.Tests/IntegrationTests/ReleaseFromHotfixTests.cs
Frazerus: That is the functionality in the test though. 
The script asks the user if he wants to create a new support branch, which is confirmed in this test on line 144. 

The test above this one oultines what happens when we do not want to create a new support branch. It merely creates the next hotfix version.

ReleaseProcessAutomation/Steps/ContinueReleaseStepWithOptionalSupportBranchStepBase.cs
```diff
@@ -0,0 +1,34 @@
+using System;
```
MichaelKetting, line 1: Missing copyright header

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #77, opened 2022-06-22
Bug/RMSRCBUILD-99 commit message of created tag is not completely written out

Frazerus:


ReleaseProcessAutomation/Git/CommandLineGitClient.cs
```diff
@@ -299,9 +299,10 @@ public void CheckoutDiscard (string fileName)
     ExecuteGitCommand($"checkout -- {fileName}");
   }
 
-  public void Tag (string arguments)
+  public void Tag (string tagName)
```
MichaelKetting, line 302: As discussed, a better soluition would be to pass both the tagname and the message as separate parameters. The GitClient should only do the commandline operation, not define (part of) the actual message.

ReleaseProcessAutomation.Tests/Git/CommandLineGitClientTests.cs
```diff
@@ -508,13 +508,24 @@ public void CheckoutDiscard_CommitOneFileAndEdit_DoesNotContainEdit ()
   }
 
   [Test]
-  public void Tag_InitCommit_FindsTagAfterwards ()
+  public void Tag_CreatesTag_FindsTagWithProperMessageAfterwards ()
```
MichaelKetting, line 511: Naming:
`Tag_WithTagNameAndMessage_CreatesTagWithProperMessage`

Operation - Parameters/Context (as needed) - Result/Expected operation

=> FindsTagAfterwards could imply that the Tag-operation does the finding. With the new name, it says that a tag should be created.

ReleaseProcessAutomation.Tests/Git/CommandLineGitClientTests.cs
```diff
@@ -508,13 +508,24 @@ public void CheckoutDiscard_CommitOneFileAndEdit_DoesNotContainEdit ()
...
 
     client.Tag("v1.0.0");
 
-    var tags = ExecuteGitCommandWithOutput("tag --merged=HEAD");
+    var tags = ExecuteGitCommandWithOutput("tag -n");
+    Assert.That(tags, Does.Contain("Create tag for version v1.0.0"));
+  }
+  
+  [Test]
+  public void Tag_CreatesTag_FindsTagAfterwards ()
```
MichaelKetting, line 522: naming / semantics. Client.Tag always creates a tag+message, => how does this test work?

MichaelKetting (changes requested):

ReleaseProcessAutomation.Tests/Git/CommandLineGitClientTests.cs
```diff
@@ -508,13 +508,24 @@ public void CheckoutDiscard_CommitOneFileAndEdit_DoesNotContainEdit ()
...
+    var tags = ExecuteGitCommandWithOutput("tag -n");
+    Assert.That(tags, Does.Contain("Create tag for version v1.0.0"));
+  }
+  
+  [Test]
+  public void Tag_TagWithoutMessage_CreatesTag ()
+  {
+    var client = new CommandLineGitClient();
+
+    client.Tag("v1.0.0", "");
```
MichaelKetting, line 526: empty string works, but `null` would be semantically cleaner.

MichaelKetting (approved):

PR #79, opened 2022-06-23
Bug/RMSRBUILD-130 merging a release branch into master sometimes causes merge conflicts

Frazerus:


MichaelKetting (changes requested):

ReleaseProcessAutomation.Tests/Git/CommandLineGitClientTests.cs
```diff
@@ -438,6 +438,62 @@ public void MergeBranch_WithoutErrors_MergesBranch ()
     logs = logs.Replace(" ", "");
     Assert.That(logs, Is.EqualTo(correctLogs));
   }
+  
+  [Test]
+  public void MergeBranch_WithConflictsAndNoAutomaticResolution_ShouldThrow ()
```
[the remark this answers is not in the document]
Frazerus: I think there are some problems here that we need to talk about, as there are some bugs and some questions I have.

Frazerus:
So about this current version: 
With commit-tree, we create a new commit based on the commithash of the `base` branch. The parents (-p) of  this commit are `HEAD` and the `base` branch, making it a merge commit.

As `commit-tree` does not do all the work that `commit` does, we need to assign the new head manually with `update-ref HEAD`.
We also need to move the `HEAD` to the new branch, done with `reset --hard`. 

There is a problem however, if we wanted use the noCommit option in this merge, we could not use the normal `commit`, therefore requiring some low-level git knowledge. This is however not intended in the current application anyways, and is possible, we would simply need to create an API for it and tell people to use ours instead of the normal git behaviour.

ReleaseProcessAutomation/Steps/PipelineSteps/ContinueReleaseOnMasterStep.cs
```diff
@@ -100,8 +100,7 @@ private void CreateTagAndMerge ()
 
     GitClient.Checkout("master");
 
-    GitClient.MergeBranch(currentBranchName);
-    GitClient.ResolveMergeConflicts();
+    GitClient.MergeBranch(currentBranchName, false, true);
```
MichaelKetting: Did you intentionally not use the named parameters or was this an oversight? If first, I would like to better understand your thoughts on this subject.

ReleaseProcessAutomation/Git/CommandLineGitClient.cs
```diff
@@ -245,10 +245,11 @@ public string CheckoutNewBranch (string branchName)
     return checkout.Output;
   }
 
-  public void MergeBranch (string branchName, bool noCommit = false)
+  public void MergeBranch (string branchName, bool noCommit = false, bool fixConflicts = false)
   {
     var shouldCommit = noCommit ? "--no-commit" : "";
-    var merge = ExecuteGitCommandWithOutput($"merge {branchName} --no-ff {shouldCommit}");
+    var shouldFixConflicts = fixConflicts ? "-Xtheirs" : "";
+    var merge = ExecuteGitCommandWithOutput($"merge {branchName} --no-ff {shouldCommit} {shouldFixConflicts}");
```
MichaelKetting: Am not sure about the latest, but shouldn't there be more logic now with the merge conflict resolution for added/removed files?

Frazerus:
There is definetly some problem here, parts of the files you commented should already be completely different

ReleaseProcessAutomation/Steps/PipelineSteps/ContinueReleaseOnMasterStep.cs
Frazerus: Just noticed, that in my working directory and the Files changed tab I already have this part with named parameters. WTF GitHub

ReleaseProcessAutomation/Git/CommandLineGitClient.cs
```diff
@@ -245,9 +245,14 @@ public string CheckoutNewBranch (string branchName)
...
   {
     var shouldCommit = noCommit ? "--no-commit" : "";
+    if (fixConflicts)
+    {
+      MergeBranchWithTheirsStrategy(branchName, noCommit);
+      return;
+    }
     var merge = ExecuteGitCommandWithOutput($"merge {branchName} --no-ff {shouldCommit}");
     if (!merge.Success)
     {
```
MichaelKetting, line 258: Code structure:
```
if (fixConflicts)
{
}
else
{
  var shouldCOmmit = ...;
  var merge = ...;
  if (!merge.Success...)
  {
  ...
  }
}
```

Reasoning: those two paths are equal. 

Actually, given the flag, we could consider splitting those into separate methods since the boolean flag seems to split things down the middle. Only makes sense if a) the public behavior is diffferent enough so good naming is possible (indicates single repsonsibility) and b) shared functionality in the future is unlikely. 

Where do we use the noCommit flag? 

Where do we use fixConflicts = false?

MichaelKetting (changes requested):

MichaelKetting:
> There is definetly some problem here, parts of the files you commented should already be completely different and are in my changelist already in the supposed-to-be state.
> Not sure what is going on here

let's look at this live.

PR #81, opened 2022-06-27
bug/RMSRCBUILD-285 DoesNotMergeIntoDevelopTest fails because of test setup

Frazerus:
This is something where we tried to improve readability, but as merge branch calls which branch should be merged into the current branch, the old version failed. 

Is there some way we can test that a specific method call was made before the current one? In that case we could check that the correct branch is checked out before merging, therefore showing the result properly again and not allowing any tests to pass that shouldnt.


MichaelKetting (commented):
See comment

PR #82, opened 2022-06-27
Feature/RMSRCBUILD-78 when not on release branch add documentation information on why no release candidate is offered

Frazerus:


ReleaseProcessAutomation.Tests/Steps/StartReleaseStepTests.cs
```diff
@@ -21,58 +21,132 @@
 using ReleaseProcessAutomation.Git;
 using ReleaseProcessAutomation.Steps;
 using ReleaseProcessAutomation.Steps.PipelineSteps;
+using Spectre.Console.Testing;
```
MichaelKetting, line 24: Please don't do this sort of reformat in during a random PR. It makes for very hard reading. Is this just the one file that's using the wrong style or the first one with the new style?

if it's the first one with the new style, it would be best to have a single commit to fix all files.

if it's the only one with the old style, also a separate commit, but one that just fixes this file.

ReleaseProcessAutomation/Steps/StartReleaseStep.cs
```diff
@@ -69,30 +73,37 @@ public void Execute (string? commitHash, bool pauseForCommit = false, bool noPus
...
       _log.Debug("On branch 'release', calling branch from release");
       _branchFromReleaseStep.Execute(commitHash, pauseForCommit, noPush);
     }
     else
     {
-      const string message = "You have to be on either a 'hotfix/*' or 'release/*' or 'develop' or 'master' branch to release a version";
-      throw new InvalidOperationException(message);
+      if(!startReleasePhase)
+        _console.WriteLine(
+            "As you are not on a release branch, you won't be able to release a release candidate version.\nTo create a release branch, use the command [green]'New-Release-Branch'[/]");
```
MichaelKetting, line 85: braces for multi-line if blocks

MichaelKetting, line 85: is the commandname a hardcoded string somewhere? if so, it would be good to use a const to prevent duplication.

ReleaseProcessAutomation/Steps/StartReleaseStep.cs
```diff
@@ -69,30 +73,37 @@ public void Execute (string? commitHash, bool pauseForCommit = false, bool noPus
...
     {
       _log.Debug("On branch 'release', calling branch from release");
       _branchFromReleaseStep.Execute(commitHash, pauseForCommit, noPush);
     }
     else
     {
-      const string message = "You have to be on either a 'hotfix/*' or 'release/*' or 'develop' or 'master' branch to release a version";
-      throw new InvalidOperationException(message);
+      if(!startReleasePhase)
+        _console.WriteLine(
```
MichaelKetting, line 84: it's like a live demo / screenshot of the new behavior, please. I think the code says that the text is printed first, not last (after the options) and I'd like to verify how it looks.

MichaelKetting (commented):
You already mered this one?

PR #83, opened 2022-06-28
Bug/RMSRCBUILD 265 release branch does not offer release candidate version as next version

Frazerus:
https://re-motion.atlassian.net/browse/RMSRCBUILD-265

### DIFF
--- a/ReleaseProcessAutomation.IntegrationTests/ReleaseFromReleaseTests.cs
+++ b/ReleaseProcessAutomation.IntegrationTests/ReleaseFromReleaseTests.cs
@@ -153,7 +153,7 @@ internal class ReleaseFromReleaseTests : IntegrationTestSetup
     //Get release version from user
     TestConsole.Input.PushTextWithEnter("1.3.5-rc.1");
     //Get next release version from user for jira
-    TestConsole.Input.PushTextWithEnter("1.3.6");
+    TestConsole.Input.PushTextWithEnter("1.3.5");
 
     var act = RunProgram(new[] { "Release-Version" });
 
@@ -227,7 +227,7 @@ internal class ReleaseFromReleaseTests : IntegrationTestSetup
     //Get release version from user
     TestConsole.Input.PushTextWithEnter("2.0.0");
     //Get next release version from user for jira
-    TestConsole.Input.PushTextWithEnter("2.1.0");
+    TestConsole.Input.PushTextWithEnter("3.0.0");
     //Do not want to create support branch
     TestConsole.Input.PushTextWithEnter("n");
 
@@ -238,7 +238,7 @@ internal class ReleaseFromReleaseTests : IntegrationTestSetup
   }
 
   [Test]
-  public void ReleaseWithRC_FromDevelopRelease_WithDevelopNotAheadOfRoot_ReleasesToMaster ()
+  public void ReleaseWithRC_FromDevelopReleaseWithMinorNextVersion_ReleasesToMaster ()
   {
     var correctLogs =
         @"*    (tag: v2.0.0, origin/master, master)Merge branch 'release/v2.0.0'
@@ -259,6 +259,37 @@ internal class ReleaseFromReleaseTests : IntegrationTestSetup
     TestConsole.Input.PushTextWithEnter("2.0.0");
     //Get next release version from user for jira
     TestConsole.Input.PushTextWithEnter("2.1.0");
+    //Do not want to create support branch
+    TestConsole.Input.PushTextWithEnter("n");
+
+    var act = RunProgram(new[] { "Release-Version" });
+
+    Assert.That(act, Is.EqualTo(0));
+    AssertValidLogs(correctLogs);
+  }
+  
+  [Test]
+  public void ReleaseWithRC_FromDevelopReleaseWithMajorNextVersion_ReleasesToMaster ()
+  {
+    var correctLogs =
+        @"*    (tag: v2.0.0, origin/master, master)Merge branch 'release/v2.0.0'
+          |\  
+          | *  (origin/release/v2.0.0, release/v2.0.0)Update metadata to version '2.0.0'.
+          |/  
+          *  (HEAD -> develop, origin/develop)feature
+          * ConfigAndBuildProject
+          * Initial CommitAll
+          ";
+
+    ExecuteGitCommand("commit -m feature --allow-empty");
+    ExecuteGitCommand("checkout -b develop");
+
+    ExecuteGitCommand("checkout -b release/v2.0.0");
+
+    //Get release version from user
+    TestConsole.Input.PushTextWithEnter("2.0.0");
+    //Get next release version from user for jira
+    TestConsole.Input.PushTextWithEnter("3.0.0");
     //Do not want to create support branch
     TestConsole.Input.PushTextWithEnter("n");
 
--- a/ReleaseProcessAutomation.Tests/Steps/Releases/ReleaseRcStepTests.cs
+++ b/ReleaseProcessAutomation.Tests/Steps/Releases/ReleaseRcStepTests.cs
@@ -1,0 +1,231 @@
+﻿// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+//
+// See the NOTICE file distributed with this work for additional information
+// regarding copyright ownership.  rubicon licenses this file to you under 
+// the Apache License, Version 2.0 (the "License"); you may not use this 
+// file except in compliance with the License.  You may obtain a copy of the 
+// License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software 
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT 
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the 
+// License for the specific language governing permissions and limitations
+// under the License.
+//
+
+using System;
+using System.IO;
+using Moq;
+using NUnit.Framework;
+using ReleaseProcessAutomation.Configuration;
+using ReleaseProcessAutomation.Configuration.Data;
+using ReleaseProcessAutomation.Extensions;
+using ReleaseProcessAutomation.Git;
+using ReleaseProcessAutomation.ReadInput;
+using ReleaseProcessAutomation.Scripting;
+using ReleaseProcessAutomation.SemanticVersioning;
+using ReleaseProcessAutomation.Steps.PipelineSteps;
+using Spectre.Console;
+
+namespace ReleaseProcessAutomation.Tests.Steps.Releases
+{
+  [TestFixture]
+  internal class ReleaseRcStepTests
+  {
+    [SetUp]
+    public void Setup ()
+    {
+      _gitClientStub = new Mock<IGitClient>();
+      _inputReaderMock = new Mock<IInputReader>();
+      _ancestorStub = new Mock<IAncestorFinder>();
+      _msBuildInvokerMock = new Mock<IMSBuildCallAndCommit>();
+      _continueAlphaBetaMock = new Mock<IContinueAlphaBetaStep>();
+
+      _consoleStub = new Mock<IAnsiConsole>();
+
+      var path = Path.Join(TestContext.CurrentContext.TestDirectory, c_configFileName);
+      _config = new ConfigReader().LoadConfig(path);
+
+    }
+
+    private Mock<IAnsiConsole> _consoleStub;
+    private Mock<IGitClient> _gitClientStub;
+    private Mock<IInputReader> _inputReaderMock;
+    private Configuration.Data.Config _config;
+    private Mock<IAncestorFinder> _ancestorStub;
+    private Mock<IMSBuildCallAndCommit> _msBuildInvokerMock;
+    private Mock<IContinueAlphaBetaStep> _continueAlphaBetaMock;
+    private const string c_configFileName = "ReleaseProcessScript.Test.Config";
+
+    [Test]
+    public void Execute_callsNextVersionFromDevelop_WithProperVersionCollection ()
+    {
+      _gitClientStub.Setup(_ => _.IsWorkingDirectoryClean()).Returns(true);
+      _gitClientStub.Setup(_ => _.IsCommitHash("")).Returns(true);
+      _gitClientStub.Setup(_ => _.IsOnBranch("release/")).Returns(true);
+      _gitClientStub.Setup(_ => _.GetCurrentBranchName()).Returns("branchName");
+      _gitClientStub.Setup(_ => _.DoesBranchExist("prerelease/v0.0.0")).Returns(false);
+      var nextPossibleVersions = new SemanticVersion().GetNextPossibleVersionsDevelop();
+
+      _inputReaderMock.Setup(_ => _.ReadVersionChoice(It.IsAny<string>(), nextPossibleVersions)).Returns(new SemanticVersion()).Verifiable();
+
+      var rcStep = new ReleaseRCStep(
+          _gitClientStub.Object,
+          _config,
+          _inputReaderMock.Object,
+          _ancestorStub.Object,
+          _msBuildInvokerMock.Object,
+          _continueAlphaBetaMock.Object,
+          _consoleStub.Object);
+
+      Assert.That(
+          () => rcStep.Execute(new SemanticVersion(), "", false, false, "develop"),
+          Throws.Nothing);
+      _inputReaderMock.Verify();
+
+    }
+
+    [Test]
+    public void Execute_callsNextVersionFromRelease_WithProperVersionCollection ()
+    {
+      _gitClientStub.Setup(_ => _.IsWorkingDirectoryClean()).Returns(true);
+      _gitClientStub.Setup(_ => _.IsCommitHash("")).Returns(true);
+      _gitClientStub.Setup(_ => _.IsOnBranch("release/")).Returns(true);
+      _gitClientStub.Setup(_ => _.GetCurrentBranchName()).Returns("branchName");
+      _gitClientStub.Setup(_ => _.DoesBranchExist("prerelease/v0.0.0")).Returns(false);
+      var nextPossibleVersions = new SemanticVersion().GetNextPossibleVersionsDevelop();
+
+      _inputReaderMock.Setup(_ => _.ReadVersionChoice(It.IsAny<string>(), nextPossibleVersions)).Returns(new SemanticVersion()).Verifiable();
+
+      var rcStep = new ReleaseRCStep(
+          _gitClientStub.Object,
+          _config,
+          _inputReaderMock.Object,
+          _ancestorStub.Object,
+          _msBuildInvokerMock.Object,
+          _continueAlphaBetaMock.Object,
+          _consoleStub.Object);
+
+      Assert.That(
+          () => rcStep.Execute(new SemanticVersion(), "", false, false, "release/v1.3.5"),
+          Throws.Nothing);
+      _inputReaderMock.Verify();
+
+    }
+
+    [Test]
+    public void Execute_callsNextVersioFromHotfix_WithProperVersionCollection ()
+    {
+      _gitClientStub.Setup(_ => _.IsWorkingDirectoryClean()).Returns(true);
+      _gitClientStub.Setup(_ => _.IsCommitHash("")).Returns(true);
+      _gitClientStub.Setup(_ => _.IsOnBranch("release/")).Returns(true);
+      _gitClientStub.Setup(_ => _.GetCurrentBranchName()).Returns("branchName");
+      _gitClientStub.Setup(_ => _.DoesBranchExist("prerelease/v0.0.0")).Returns(false);
+      var nextPossibleVersions = new SemanticVersion().GetNextPossibleVersionsHotfix();
+
+      _inputReaderMock.Setup(_ => _.ReadVersionChoice(It.IsAny<string>(), nextPossibleVersions)).Returns(new SemanticVersion()).Verifiable();
+
+      var rcStep = new ReleaseRCStep(
+          _gitClientStub.Object,
+          _config,
+          _inputReaderMock.Object,
+          _ancestorStub.Object,
+          _msBuildInvokerMock.Object,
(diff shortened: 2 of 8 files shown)


ReleaseProcessAutomation.Tests/SemanticVersioning/SemanticVersionTests.cs
```diff
@@ -46,6 +46,25 @@ public void GetNextPossibleVersionsDevelop_WithPreReleaseAlpha_ReturnsSixVersion
     Assert.That(possibleVersions, Does.Contain(parser.ParseVersion("1.3.5-beta.1")));
   }
 
+  [Test]
+  public void GetNextPossibleVersionsDevelop_WithPreReleaseRC_ReturnsProperVersions()
```
MichaelKetting, line 50: The test-name refers to develop-branch. the RC-version should only be on release-branches. When you're on develop, you get the info-text that you need to make a release-branch for RC versions. 

This seems to also affect the helper-method `GetNextPossibleVersionsDevelop`.

MichaelKetting (changes requested):

Frazerus:
Please put special attention on whether the new methods create the proper versions as I'm not 100% sure that they do.

ReleaseProcessAutomation/Steps/PipelineSteps/ReleaseWithRcStep.cs
```diff
@@ -100,12 +100,12 @@ public void Execute (bool pauseForCommit, bool noPush, string ancestor)
     if (ancestor.Equals("develop"))
     {
       _log.Debug("Getting next possible jira versions for develop from '{nextVersion}'", nextVersion);
```
MichaelKetting, line 102: Likely a follow-up ticket: There's two steps `ReleaseWithRcStep` and ReleaseRCStep`. 
* casing in names
* what's the difference between those types? The docs seem similar, the naming, too, and this branch-logic here as well?.
* The one difference is that one type checks just develop and the other develop and release-branch. 

=> maybe just needs better naming? Maybe there's another problem, too?

ReleaseProcessAutomation/Extensions/SemanticVersionExtensions.cs
```diff
@@ -194,6 +226,15 @@ public static SemanticVersion GetNextPatchVersion (this SemanticVersion semantic
     return nextPatchVersion;
   }
 
+  private static IReadOnlyCollection<SemanticVersion> ReleaseRcOrCurrent (SemanticVersion semanticVersion)
+  {
+    return new[]
+           {
+               GetNextRc(semanticVersion),
```
MichaelKetting, line 233: Casing: When there's a two-letter acronym, it's uppercase. Either way though, the current code base is inconsistent. We should either do all-upper-case naming for `RC` or make the code base use lower-case if that's a) what we want and b) is less effort because there's way less upper-case spellings in the code.

ReleaseProcessAutomation/Extensions/SemanticVersionExtensions.cs
```diff
@@ -26,6 +26,36 @@ public static class SemanticVersionExtensions
...
+    {
+      return ReleaseRcOrCurrent(semanticVersion);
+    }
+    return new[]
+           {
+               GetNextMajorWithAlpha(semanticVersion),
+               GetNextMajorWithBeta(semanticVersion),
+               GetNextMajorRelease(semanticVersion)
+           };
+  }
```
MichaelKetting, line 41: I'm not 100% sure about the semantics of this method. When I do a non-rc branch, I should get the alpha/beta for the current rtm version (major.minor.0), not "major.0.0" and I think the last option GetNextMajorRelease is also not current. 

It's probably something that's a hold over from the open review of everything. Maybe we just need to talk this through?

MichaelKetting (changes requested):
needs clenaring up of the open questions first

ReleaseProcessAutomation/Extensions/SemanticVersionExtensions.cs
Frazerus: Created a task for it

ReleaseProcessAutomation/Steps/PipelineSteps/ReleaseWithRcStep.cs
Frazerus: Created a task for evaluation of uses and naming

ReleaseProcessAutomation/Extensions/SemanticVersionExtensions.cs
Frazerus: Yeah, quick live sessions seem to be in order

a-ctor (changes requested):
@Frazerus Please rebase this PR as it is very old and has a few merge conflicts. Re-request the review once rebased and I will take a look at it :)

a-ctor (changes requested):

ReleaseProcessAutomation.Tests/Steps/Releases/ReleaseRcStepTests.cs
```diff
@@ -0,0 +1,231 @@
...
+    private Mock<IGitClient> _gitClientStub;
+    private Mock<IInputReader> _inputReaderMock;
+    private Configuration.Data.Config _config;
+    private Mock<IAncestorFinder> _ancestorStub;
+    private Mock<IMSBuildCallAndCommit> _msBuildInvokerMock;
+    private Mock<IContinueAlphaBetaStep> _continueAlphaBetaMock;
+    private const string c_configFileName = "ReleaseProcessScript.Test.Config";
+
+    [Test]
+    public void Execute_callsNextVersionFromDevelop_WithProperVersionCollection ()
```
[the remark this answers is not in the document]
Frazerus: This is happening in a different issue alltogether as there are several spots where the test names are questionable at best. Will edit the ones here though, as I'm already into it.

a-ctor (approved):

ReleaseProcessAutomation.Tests/Steps/Releases/ReleaseRcStepTests.cs
```diff
@@ -0,0 +1,231 @@
...
+          _inputReaderMock.Object,
+          _ancestorStub.Object,
+          _msBuildInvokerMock.Object,
+          _continueAlphaBetaMock.Object,
+          _consoleStub.Object);
+
+      Assert.That(
+          () => rcStep.Execute(new SemanticVersion(), "", false, false, "hotfix/v1.3.5"),
+          Throws.Nothing);
+      _msBuildInvokerMock.Verify();
```
MichaelKetting, line 228: It's not wrong to do it this way, but you could just call the Execute method and then mock.Verify. No need to assert that no exception is thrown since this would fail the test anyway and the mock.Verify makes it clear what we are asserting.

ReleaseProcessAutomation.Tests/Steps/Releases/ReleaseRcStepTests.cs
```diff
@@ -0,0 +1,231 @@
...
+    [Test]
+    public void Execute_WithoutErrors_CallsMBuildInvokeAndCommit()
+    {
+      _gitClientStub.Setup(_ => _.IsWorkingDirectoryClean()).Returns(true);
+      _gitClientStub.Setup(_ => _.IsCommitHash("")).Returns(true);
+      _gitClientStub.Setup(_ => _.IsOnBranch("release/")).Returns(true);
+      _gitClientStub.Setup(_ => _.GetCurrentBranchName()).Returns("branchName");
+      _gitClientStub.Setup(_ => _.DoesBranchExist("prerelease/v0.0.0")).Returns(false);
+      var nextPossibleVersions = new SemanticVersion().GetNextPossibleVersionsHotfix();
+      _msBuildInvokerMock.Setup(_ => _.CallMSBuildStepsAndCommit(MSBuildMode.PrepareNextVersion, new SemanticVersion())).Verifiable();
```
MichaelKetting, line 213: Stylistic choice: if you have only void-typed calls that you wish to verify, you can move the verification to the bottom of the test, using the AAA style (arrange, act, assert). It's okay to mix these styles between tests, but you should avoid mixing them within the same test. 

Of course, if the tests vary only in their verifiable setups and some use void and others have a return statement, it might help readability of the entire fixture, to stick to verifable throughout.  => I did not check the entire fixture if that's the case, but it seems like most (all?) tests use void + verifiable.

ReleaseProcessAutomation.Tests/Steps/Releases/ReleaseRcStepTests.cs
```diff
@@ -0,0 +1,231 @@
...
+    public void Execute_WithoutErrors_CallsNextStep ()
+    {
+      _gitClientStub.Setup(_ => _.IsWorkingDirectoryClean()).Returns(true);
+      _gitClientStub.Setup(_ => _.IsCommitHash("")).Returns(true);
+      _gitClientStub.Setup(_ => _.IsOnBranch("release/")).Returns(true);
+      _gitClientStub.Setup(_ => _.GetCurrentBranchName()).Returns("branchName");
+      _gitClientStub.Setup(_ => _.DoesBranchExist("prerelease/v0.0.0")).Returns(false);
+      var nextPossibleVersions = new SemanticVersion().GetNextPossibleVersionsHotfix();
+      _continueAlphaBetaMock.Setup(_ => _.Execute(It.IsAny<SemanticVersion>(), It.IsAny<string>(), It.IsAny<string>(), false)).Verifiable();
+
```
MichaelKetting, line 186: drop blank line?

ReleaseProcessAutomation.Tests/Steps/Releases/ReleaseRcStepTests.cs
```diff
@@ -0,0 +1,231 @@
...
+    [Test]
+    public void Execute_callsNextVersioFromHotfix_WithProperVersionCollection ()
+    {
+      _gitClientStub.Setup(_ => _.IsWorkingDirectoryClean()).Returns(true);
+      _gitClientStub.Setup(_ => _.IsCommitHash("")).Returns(true);
+      _gitClientStub.Setup(_ => _.IsOnBranch("release/")).Returns(true);
+      _gitClientStub.Setup(_ => _.GetCurrentBranchName()).Returns("branchName");
+      _gitClientStub.Setup(_ => _.DoesBranchExist("prerelease/v0.0.0")).Returns(false);
+      var nextPossibleVersions = new SemanticVersion().GetNextPossibleVersionsHotfix();
+
```
MichaelKetting, line 127: drop blank line

ReleaseProcessAutomation.Tests/Steps/Releases/ReleaseRcStepTests.cs
```diff
@@ -0,0 +1,231 @@
...
+    [Test]
+    public void Execute_callsNextVersioFromHotfix_WithProperVersionCollection ()
+    {
+      _gitClientStub.Setup(_ => _.IsWorkingDirectoryClean()).Returns(true);
+      _gitClientStub.Setup(_ => _.IsCommitHash("")).Returns(true);
+      _gitClientStub.Setup(_ => _.IsOnBranch("release/")).Returns(true);
+      _gitClientStub.Setup(_ => _.GetCurrentBranchName()).Returns("branchName");
+      _gitClientStub.Setup(_ => _.DoesBranchExist("prerelease/v0.0.0")).Returns(false);
+      var nextPossibleVersions = new SemanticVersion().GetNextPossibleVersionsHotfix();
```
MichaelKetting, line 126: maybe add blank line here?

ReleaseProcessAutomation.Tests/Steps/Releases/ReleaseRcStepTests.cs
MichaelKetting, line 63: please check other methods on this fixtures for similar occurances of the recommended changes

ReleaseProcessAutomation.IntegrationTests/ReleaseFromReleaseTests.cs
```diff
@@ -227,7 +227,7 @@ public void ReleaseRelease_FromDevelopRelease_WithDevelopAheadOfRoot_ReleasesToM
     //Get release version from user
     TestConsole.Input.PushTextWithEnter("2.0.0");
     //Get next release version from user for jira
-    TestConsole.Input.PushTextWithEnter("2.1.0");
+    TestConsole.Input.PushTextWithEnter("3.0.0");
```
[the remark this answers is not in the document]
Frazerus: We should not be releasing that version to master though, right? The jira version does not matter for this test, as this runs without using jira anyways as that would be risky and a ginormous overhead. This and the next are identical though, so we should probably remove one of them.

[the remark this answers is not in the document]
MichaelKetting: if/why the version matters: if there's non-jira logic based on the version, or if the list of versions is printed out.

and yes, we do can do such a release. for instance, remotion Framework does it right now. The just-major variant is actually the special case.

ReleaseProcessAutomation.UnitTests/Steps/Releases/ReleaseRcStepTests.cs
```diff
@@ -104,8 +105,8 @@ public void Execute_callsNextVersionFromRelease_WithProperVersionCollection ()
     _gitClientStub.Setup(_ => _.IsOnBranch("release/")).Returns(true);
     _gitClientStub.Setup(_ => _.GetCurrentBranchName()).Returns("branchName");
     _gitClientStub.Setup(_ => _.DoesBranchExist("prerelease/v0.0.0")).Returns(false);
-    var nextPossibleVersions = nextVersion.GetNextPossibleVersionsDevelop();
-
+    
+    var nextPossibleVersions = nextVersion.GetNextPossibleVersionsReleaseBranchFromDevelop();
```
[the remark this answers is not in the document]
Frazerus: Shouldn't the existing logic be tested in a separate unit test anyways? This should only check if it is using the correct extension method, if there is a bug with the extension method it should already be detected by the corresponding unit tests. 

Since we can't verify the correct call by using mocks, this is IMO the best option we have.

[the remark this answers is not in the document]
MichaelKetting: yes and no (about "shouldn't it be tested anway?"). Any value that can be statically derived from the testsetup (and without the use of pen and paper and a one-time-pad) should still be hardcoded. Here, for instance, it seems from reading the test (not checked the prod-code) we're using the same method calls in both the test and production. This is typically a case of code duplication which could result in copy&paste errors (doing the wrong method call and copying it into the test as well) or other effects. It also hides what we actually try to do because you don't see the relevant information in plain string form but rather obfuscated behind API calls. 

For instance, if you do a call to string.Trim() or a call to string.Split(), the APIs you use are probably quite bugfree, but you can't be certain about the actual semantics. For instance, with Split(), you might pass a parameter that's incorrect and you don't notice because you have the same mistake in both places.

That said, I've looked at the test again and realize now, it's actually a list of versions you're getting, not just a single version and this result is used as a parameter in the setup-call. And the list might change in the future. Okay, yes, it's okay to do it like this here because inlining the result would pose a significant risk of future changes. Thanks for challangening :)

It also shows that the abstraction is possibly not perfect, a mockable service instead of an API on SemVersion might be the better choice. It's what you call "in hindsight" :D

ReleaseProcessAutomation/Steps/PipelineSteps/ReleaseRCStep.cs
```diff
@@ -93,12 +93,12 @@ public void Execute (SemanticVersion nextVersion, string? commitHash, bool pause
     if (ancestor.Equals("develop") || ancestor.StartsWith("release/"))
```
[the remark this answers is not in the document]
Frazerus: Hmm, well ReleaseWithRC basically means to release from a release branch to a full version, while ReleaseRC just means to release a new RC. Therefore I would suggest:
ReleaseRC -> ReleaseRCOnReleaseBranch
ReleaseWithRC -> ReleaseFullVersionFromReleaseBranch

There is already a ticket for this: https://re-motion.atlassian.net/browse/RMRPA-4

ReleaseProcessAutomation.UnitTests/SemanticVersioning/SemanticVersionTests.cs
```diff
@@ -173,6 +252,73 @@ public void GetCurrentPossibleVersionsHotfix_WithPreReleaseRC_ReturnsVersions ()
...
+    Assert.That(possibleVersions, Is.EqualTo(expectedVersions));
+  }
+  
+  [Test]
+  public void GetNextPossibleVersionsReleaseBranchFromDevelop_WithFullRelease_ReturnsExpectedVersions ()
+  {
+    var parser = new SemanticVersionParser();
+    var expectedVersions = new SemanticVersion[]
+                           {                               
+                               parser.ParseVersion("2.1.0"),
```
MichaelKetting, line 278: should we add 2.1.0-alpha and beta, too?

and a separate topic (ticket): we should also add logic, that the associated release-version is added implicitly to jira, when a prerelease version is selected. i.e. it's not helpful to decide we want a 2.1.0-alpha.1 when there's not also a 2.1.0 version in jira.

ReleaseProcessAutomation.UnitTests/SemanticVersioning/SemanticVersionTests.cs
```diff
@@ -173,6 +252,73 @@ public void GetCurrentPossibleVersionsHotfix_WithPreReleaseRC_ReturnsVersions ()
...
+                           };
+    var baseVersion = parser.ParseVersion("1.3.5-rc.1");
+
+    var possibleVersions = baseVersion.GetNextPossibleVersionsForReleaseBranchFromDevelop();
+
+    Assert.That(possibleVersions, Is.EqualTo(expectedVersions));
+  }
+  
+  [Test]
+  public void GetNextPossibleVersionsReleaseBranchFromDevelop_WithFullRelease_ReturnsExpectedVersions ()
```
MichaelKetting, line 273: I think we also want a test for minor version, see GetNextPossibleVersionsDevelop_WithMinorVersion_ReturnsExpectedVersions and GetNextPossibleVersionsDevelop_WithMajorVersion_ReturnsExpectedVersions

ReleaseProcessAutomation.UnitTests/SemanticVersioning/SemanticVersionTests.cs
```diff
@@ -42,7 +42,86 @@ public void GetNextPossibleVersionsDevelop_WithPreReleaseAlpha_ReturnsSixVersion
...
+                           };
+    var baseVersion = parser.ParseVersion("1.3.0-rc.1");
+
+    var possibleVersions = baseVersion.GetNextPossibleVersionsDevelop();
+
+    Assert.That(possibleVersions, Is.EqualTo(expectedVersions));
+  }
+  
+  [Test]
+  public void GetNextPossibleVersionsDevelop_WithMajorVersionWithoutPreRelease_ReturnsExpectedVersions()
```
MichaelKetting, line 64: I can't spot the difference between this test and GetNextPossibleVersionsDevelop_WithMajorVersion_ReturnsExpectedVersions

(i.e. why is the list of expected versions different, the rest seems identical to me. please spot the difference for me ;) )

ReleaseProcessAutomation.Tests/Steps/Releases/ReleaseRcStepTests.cs
```diff
@@ -0,0 +1,231 @@
...
+    [Test]
+    public void Execute_WithPauseForCommit_DoesNotCallNextStep ()
+    {
+      _gitClientStub.Setup(_ => _.IsWorkingDirectoryClean()).Returns(true);
+      _gitClientStub.Setup(_ => _.IsCommitHash("")).Returns(true);
+      _gitClientStub.Setup(_ => _.IsOnBranch("release/")).Returns(true);
+      _gitClientStub.Setup(_ => _.GetCurrentBranchName()).Returns("branchName");
+      _gitClientStub.Setup(_ => _.DoesBranchExist("prerelease/v0.0.0")).Returns(false);
+      var nextPossibleVersions = new SemanticVersion().GetNextPossibleVersionsHotfix();
+      _continueAlphaBetaMock.Setup(_ => _.Execute(It.IsAny<SemanticVersion>(), It.IsAny<string>(), It.IsAny<string>(), false)).Verifiable();
```
MichaelKetting, line 155: this is no redundant sicne we already verify below. Do you want to use the AAA syntax in this file or stick to setup + verify. right now, you only did 1/2 of the change in this one test so I'm a bit confused.

MichaelKetting (changes requested):

ReleaseProcessAutomation.UnitTests/SemanticVersioning/SemanticVersionTests.cs
Frazerus: `GetNextPossibleVersionsDevelop(bool withoutPreRelease = false)`
We are testing the difference between having the method suggest all versions or only the ones that aren't prereleases.

ReleaseProcessAutomation.Tests/Steps/Releases/ReleaseRcStepTests.cs
Frazerus: Oh no, this isnt the right file. There are some leftovers from renaming all the files, this is going to be an extra step when swapping to the new repo. Added it in the description: https://re-motion.atlassian.net/browse/RMRPA-6

ReleaseProcessAutomation.UnitTests/SemanticVersioning/SemanticVersionTests.cs
Frazerus: Created https://re-motion.atlassian.net/browse/RMRPA-11

ReleaseProcessAutomation.Tests/Steps/Releases/ReleaseRcStepTests.cs
Frazerus: All ReleaseProcessAutomation.Tests.*** Files can be ignored

MichaelKetting (approved):

PR #86, opened 2022-06-28
Feature/RMSRCBUILD-134 check if configured ms build path exists

Frazerus:


ReleaseProcessAutomation/Scripting/MSBuildCallAndCommit.cs
```diff
@@ -57,7 +58,14 @@ public int CallMSBuildStepsAndCommit (MSBuildMode msBuildMode, SemanticVersion v
...
       _log.Warning("No MSBuild Path specified in config, will continue without MSBuild");
-      _console.WriteLine("There was no MSBuildPath specified in the config, will continue without Invoking MSBuild");
+      _console.WriteLine("There was no MSBuildPath specified in the config\nWill continue without invoking MSBuild");
+      return -1;
+    }
+
+    if (!File.Exists(msBuildPath))
+    {
+      _log.Warning("The configured MSBuildpath does not exist");
+      _console.WriteLine("The configured MSBuildPath does not exist\nPlease configure a proper MSBuildPath in the config\nWill continue without invoking MSBuild");
```
MichaelKetting, line 68: Please print the msbuild path in both the console and the log message, otherwise you have to dig and reverse-engineer if you think your path is correct, but isn't. Use single-quotes for delimit the path value in the message or place it in parenthesis. (i.e. the style of the delimiter depends on how you format the message)

ReleaseProcessAutomation/Scripting/MSBuildCallAndCommit.cs
```diff
@@ -57,7 +58,14 @@ public int CallMSBuildStepsAndCommit (MSBuildMode msBuildMode, SemanticVersion v
...
     {
       _log.Warning("No MSBuild Path specified in config, will continue without MSBuild");
-      _console.WriteLine("There was no MSBuildPath specified in the config, will continue without Invoking MSBuild");
+      _console.WriteLine("There was no MSBuildPath specified in the config\nWill continue without invoking MSBuild");
+      return -1;
+    }
+
+    if (!File.Exists(msBuildPath))
+    {
+      _log.Warning("The configured MSBuildpath does not exist");
```
MichaelKetting, line 67: please use `.` / full-stop for sentences. Most log-messages are (should be) complete sentences.

MichaelKetting (changes requested):

ReleaseProcessAutomation.Tests/MsBuild/MsBuildInvokerTests.cs
```diff
@@ -59,15 +60,77 @@ public void InvokeMSBuildAndCommit_NoMSBuildPathSpecified_EndsEarly()
...
+
+        var testConsole = new TestConsole();
+
+        var msBuildInvoker = new MSBuildCallAndCommit(gitClientStub.Object, _config, msBuildMock.Object,
+                testConsole);
+
+        var act = msBuildInvoker.CallMSBuildStepsAndCommit(MSBuildMode.PrepareNextVersion, version);
+
+        Assert.That(act, Is.EqualTo(-1));
+        Assert.That(testConsole.Output.ReplaceLineEndings(""), Does.Contain($"The configured MSBuildPath '{_config.MSBuildSettings.MSBuildPath}' does not exist"));
```
MichaelKetting, line 131: Ah okay. Yes. we need the absolute path in the output, otherwise errors due to a mis-matched Env.CurrentDirectory can drive you inside. (talking from experience here :) )

ReleaseProcessAutomation/Scripting/MSBuildCallAndCommit.cs
```diff
@@ -56,8 +57,15 @@ public int CallMSBuildStepsAndCommit (MSBuildMode msBuildMode, SemanticVersion v
...
-      _log.Warning("No MSBuild Path specified in config, will continue without MSBuild");
-      _console.WriteLine("There was no MSBuildPath specified in the config, will continue without Invoking MSBuild");
+      _log.Warning("No MSBuild Path specified in config, will continue without MSBuild.");
+      _console.WriteLine("There was no MSBuildPath specified in the config.\nWill continue without invoking MSBuild.");
+      return -1;
+    }
+
+    if (!File.Exists(msBuildPath))
+    {
+      _log.Warning("The configured MSBuildpath '{MSBuildpath}' does not exist.", msBuildPath);
```
MichaelKetting, line 67: maybe it's a syntax-highlight thingy, but there's `{MSBuildPath}` and `msBuildPath` in the code. shouldn't those be the same variable/field/property?

MichaelKetting (changes requested):

ReleaseProcessAutomation/Scripting/MSBuildCallAndCommit.cs
Frazerus: Not sure if this works, as I think I tested it like that already. Needed more info the in the message anyways, therefore changed to enumeration

ReleaseProcessAutomation/Scripting/MSBuildCallAndCommit.cs
```diff
@@ -56,8 +57,15 @@ public int CallMSBuildStepsAndCommit (MSBuildMode msBuildMode, SemanticVersion v
...
-      _console.WriteLine("There was no MSBuildPath specified in the config, will continue without Invoking MSBuild");
+      _log.Warning("No MSBuild Path specified in config, will continue without MSBuild.");
+      _console.WriteLine("There was no MSBuildPath specified in the config.\nWill continue without invoking MSBuild.");
+      return -1;
+    }
+
+    if (!File.Exists(msBuildPath))
+    {
+      _log.Warning("The configured MSBuildpath '{1}' does not exist.Current directory: '{2}'", msBuildPath, Environment.CurrentDirectory);
+      _console.WriteLine($"The configured MSBuildPath '{msBuildPath}' does not exist.\nCurrent directory: '{Environment.CurrentDirectory}'\nPlease configure a proper MSBuildPath in the config.\nWill continue without invoking MSBuild.");
```
MichaelKetting, line 68: Ah, okay. My intention was to use Path.GetFullPath (msBuildPath) which resolves the path. Sorry that this wasn't explicit before. 

Upon further consideration: you could actually do this right at the start, when you read the msBuildPath from the config file.  This removes any effects from changing directories etc during the execution / when spawning new processes.

There's also a missing whitespace in line 67 before 'Current Directory'

MichaelKetting (changes requested):

ReleaseProcessAutomation.Tests/MsBuild/MsBuildInvokerTests.cs
```diff
@@ -59,15 +60,79 @@ public void InvokeMSBuildAndCommit_NoMSBuildPathSpecified_EndsEarly()
...
+        testConsole.Profile.Width = Int32.MaxValue;
+
+        var msBuildInvoker = new MSBuildCallAndCommit(gitClientStub.Object, _config, msBuildMock.Object,
+                testConsole);
+
+        var act = msBuildInvoker.CallMSBuildStepsAndCommit(MSBuildMode.PrepareNextVersion, version);
+
+        Assert.That(act, Is.EqualTo(-1));
+        Assert.That(testConsole.Output, Does.Contain($"The configured MSBuildPath '{_config.MSBuildSettings.MSBuildPath}' does not exist"));
+        Assert.That(testConsole.Output, Does.Contain($"Current directory: '{Environment.CurrentDirectory}'"));
```
MichaelKetting, line 133: Please update. I think the second assertion is now red?

MichaelKetting (approved):
Please fix test before merge

PR #87, opened 2022-06-28
bug/RMSRCBUILD-182 Releasing an alpja

Frazerus:


ReleaseProcessAutomation.Tests/IntegrationTests/ReleaseFromDevelopTests.cs
```diff
@@ -423,4 +423,24 @@ public void ReleaseNewBranch_FromDevelop ()
     Assert.That(TestConsole.Output, Does.Not.Contain("Called UpdateAssemblyInfosForDevelopment!"));
     Assert.That(act1, Is.EqualTo(0));
   }
+
+  [Test]
+  public void ReleaseVersion_WithMajorReleaseTagOnMaster_SuggestsAlphaVersions ()
+  {
+
```
MichaelKetting, line 430: extra whitespace found

ReleaseProcessAutomation.Tests/IntegrationTests/ReleaseFromDevelopTests.cs
```diff
@@ -423,4 +423,24 @@ public void ReleaseNewBranch_FromDevelop ()
...
+    ExecuteGitCommand("tag v2.0.0 -m v2.0.0");
+    ExecuteGitCommand("checkout -b develop");
```
MichaelKetting, line 432: I'm suprised this just works. Maybe the repro-description was mis-leading?

"repository freshly checked out" means you did just fetch all the data from the server, but you have not actually checked out (i.e. created a local version) of the respective branches.

You cannot test this without using a source repository from which to fetch your local data. If you just have a local repository, all your branches are "checked out" or they would not exist at all.

MichaelKetting (changes requested):

ReleaseProcessAutomation.Tests/IntegrationTests/ReleaseFromDevelopTests.cs
Frazerus: I'm more or less sure that i fixed this at some point during the development process of the new RPA.

MichaelKetting: which is done in the new version...

MichaelKetting (commented):
I would like to verify something in the new setup before completing this PR.:

live-step-through of lines 431-438 with a git-client next to it, so I can make sure we have the same understanding. Right now, there's a couple of git details I'm not 100% clear on.

Frazerus:
The name GitBranchOperations is very much still up for debate, but I think this is a good version to be able to use `ensureBranchUpToDate` outside of the steps. 

Considering moving other parts of the releaseProcessBase into it, as the method `MergeBranchWithReset` and `CreateNewSupportBranch` would also be a good fit.

MichaelKetting:
@a-ctor I'd like to pass this one to you. I'm a bit pressed for time right now and there's many good edits in here that need a review :)
cc @Frazerus

a-ctor:
@Frazerus Please rebase this PR as we have a few merge conflicts which will probably require further edits. Re-request the review once rebased and I will take a look at it :)

a-ctor (changes requested):
Rebase please :)

ReleaseProcessAutomation/Git/SemanticVersionedGitRepository.cs
```diff
@@ -54,6 +56,11 @@ public bool TryGetCurrentVersion ([MaybeNullWhen(false)] out SemanticVersion ver
   {
     _log.Debug("Trying to get first version from '{From}' to '{To}'.", from, to);
 
+    if(!from.Equals("HEAD"))
+      _gitBranchOperations.EnsureBranchUpToDate(from);
+    if(!string.IsNullOrEmpty(to))
+      _gitBranchOperations.EnsureBranchUpToDate(to);
```
a-ctor, line 62:
```suggestion
    if (!from.Equals("HEAD"))
      _gitBranchOperations.EnsureBranchUpToDate(from);
    if (!string.IsNullOrEmpty(to))
      _gitBranchOperations.EnsureBranchUpToDate(to);
```

ReleaseProcessAutomation.UnitTests/Steps/Continues/ContinueReleaseOnMasterStepTests.cs
```diff
@@ -100,12 +103,13 @@ public void CreateTagAndMerge_WithAlreadySetTag_ThrowsException ()
         .Returns(new SemanticVersion());
 
     var step = new ContinueReleaseOnMasterStep(
```
a-ctor, line 105: All the indents in this file were correct but are now on x2 indent. You should check your settings it seems there is some setting missing.

a-ctor (approved):
Looks good. Some formatting issues are left.

PR #89, opened 2022-06-29
Feature/RMSRCBUILD 43 list issues and ask for approval before moving them

Frazerus:
[Jira Issue](https://re-motion.atlassian.net/browse/RMSRCBUILD-43)


ReleaseProcessAutomation/Jira/ServiceFacadeImplementations/JiraProjectVersionService.cs
```diff
@@ -21,22 +21,30 @@
...
 using ReleaseProcessAutomation.SemanticVersioning;
 using RestSharp;
+using Spectre.Console;
 
 namespace ReleaseProcessAutomation.Jira.ServiceFacadeImplementations;
 
 public class JiraProjectVersionService : IJiraProjectVersionService
 {
+  private readonly IInputReader _inputReader;
+  private readonly IAnsiConsole _console;
```
a-ctor, line 34: I kind of dislike that we have to pull in the input/output into this service. Ideally the services just provide the functionality and the steps are the ones that do the interaction and invoke the right functionality. If we start crossing lines it gets messy quickly.

Maybe we can extract this into a step to keep them separated? What do you think?

a-ctor (changes requested):

ReleaseProcessAutomation.IntegrationTests/Jira/JiraProjectVersionServiceTest.cs
```diff
@@ -2,8 +2,14 @@
...
 using Moq;
 using NUnit.Framework;
+using ReleaseProcessAutomation.Configuration.Data;
+using ReleaseProcessAutomation.Jira;
 using ReleaseProcessAutomation.Jira.ServiceFacadeImplementations;
 using ReleaseProcessAutomation.Jira.Utility;
+using ReleaseProcessAutomation.ReadInput;
+using ReleaseProcessAutomation.SemanticVersioning;
+using ReleaseProcessAutomation.Steps.SubSteps;
+using Spectre.Console.Testing;
```
a-ctor, line 12: A lot of those are probably not needed?

ReleaseProcessAutomation.IntegrationTests/Jira/JiraVersionFinderTests.cs
```diff
@@ -0,0 +1,70 @@
...
+
+    // Get latest unreleased version
+    var versions = _versionFinder.FindUnreleasedVersions(c_jiraProjectKey, "4.1.").ToList();
+    Assert.That(versions.Count(), Is.EqualTo(3));
+
+    var versionToRelease = versions.First();
+    Assert.That(versionToRelease.name, Is.EqualTo("4.1.0"));
+
+    var versionToFollow = versions.Skip(1).First();
+    Assert.That(versionToFollow.name, Is.EqualTo("4.1.1"));
```
a-ctor, line 60: I think we can simplify this assert by using the `Is.EquivalentTo` constraint. If we just want to check the names we can do a select.

    Assert.That(versions.Select(e => e.name), Is.EquivalentTo(new [] { "4.1.0", "4.1.1", "4.1.2" });

ReleaseProcessAutomation.IntegrationTests/Jira/JiraVersionFinderTests.cs
```diff
@@ -0,0 +1,70 @@
...
+
+    var versions2 = _versionFinder.FindUnreleasedVersions(c_jiraProjectKey, "4.2.");
+    Assert.That(versions2.Count(), Is.EqualTo(1));
+
+    var additionalVersion = versions2.First();
+    Assert.That(additionalVersion.name, Is.EqualTo("4.2.0"));
```
a-ctor, line 66: Same here see comment above.

ReleaseProcessAutomation.IntegrationTests/Steps/JiraSubStepTests.cs
```diff
@@ -0,0 +1,192 @@
...
+    // Create versions
+    _service.CreateVersion(c_jiraProjectKey, "4.1.0", DateTime.Today.AddDays(1));
+    _service.CreateSubsequentVersion(c_jiraProjectKey, "4\\.1\\..*", 3, DayOfWeek.Monday);
+    _service.CreateSubsequentVersion(c_jiraProjectKey, "4\\.1\\..*", 3, DayOfWeek.Tuesday);
+    _service.CreateVersion(c_jiraProjectKey, "4.2.0", DateTime.Today.AddDays(7));
+
+    // Get latest unreleased version
+    var versions = _versionFinder.FindUnreleasedVersions(c_jiraProjectKey, "4.1.").ToList();
+
+    var versionToRelease = versions.First();
```
a-ctor, line 80: For the readability of the test it might be better to explicitly mention the version that we want to release. This way, it allows the reader to quickly gather what the value of the variable is. This also has the added benefit of throwing an exception when the version numbers in the test change for any reason.

```suggestion
    var versionToRelease = versions.First(e => e.name == "4.1.0");
```

ReleaseProcessAutomation.IntegrationTests/Steps/JiraSubStepTests.cs
```diff
@@ -0,0 +1,192 @@
...
+
+    var additionalVersion = _versionFinder.FindUnreleasedVersions(c_jiraProjectKey, "4.2.").First();
```
a-ctor, line 82: Here we could also consider using `.Single` instead of `.First`. The difference is that `.Single` throws if there are more than one element in enumerable. In this case it would add the implicit assertion that we expect only one version with the "4.2." prefix.

I will stop commenting on the issue but there are a few more cases of this in this file

ReleaseProcessAutomation.UnitTests/Steps/JiraSubStepTests.cs
```diff
@@ -0,0 +1,118 @@
...
+    var jiraSubStep = new JiraSubStep(
+        _console,
+        _inputReaderMock.Object,
+        _jiraIssueServiceMock.Object,
+        _jiraVersionCreatorMock.Object,
+        _jiraReleaserMock.Object);
+
+    Assert.That(() => jiraSubStep.Execute(currentVersion, nextVersion, true), Throws.Nothing);
+
+    _jiraVersionCreatorMock.Verify();
```
a-ctor, line 64: Looking at this test I wondered why it would succeed but loose mock are apparently very loose in their behavior ^^

It made me realize two things:
 - I think we should make the input explicit (strict mock) and also check the output
 - If there are no issues to move we probably don't need to ask the user if they want to move issues

a-ctor (changes requested):

ReleaseProcessAutomation.IntegrationTests/IntegrationTestSetup.cs
```diff
@@ -83,16 +84,19 @@ protected void AssertValidLogs (string expectedLogs1, string expectedLogs2)
   protected int RunProgram (string[] args)
   {
     var services = new ApplicationServiceCollectionFactory().CreateServiceCollection();
-    var jiraFunctionalityStub = new Mock<IJiraFunctionality>();
-    var jiraFunctionalityDescriptor = new ServiceDescriptor(typeof(IJiraFunctionality), x => jiraFunctionalityStub.Object, ServiceLifetime.Singleton);
+    var jiraSubStepStub = new Mock<IReleaseVersionAndMoveIssuesSubStep>();
```
a-ctor, line 87: Leftover `SubStep`

a-ctor (approved):
One rename left and then squash the commits down and merge :)

PR #90, opened 2022-06-29
Feature/RMSRCBUILD-138 correct jira fix version when creating an alpha/beta build when no corresponding jira version exists

Frazerus:
[Jira issue](https://re-motion.atlassian.net/browse/RMSRCBUILD-138)

### DIFF
--- a/ReleaseProcessAutomation/Jira/JiraVersionCreator.cs
+++ b/ReleaseProcessAutomation/Jira/JiraVersionCreator.cs
@@ -16,6 +16,7 @@
 //
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using ReleaseProcessAutomation.Configuration.Data;
 using ReleaseProcessAutomation.Jira.ServiceFacadeImplementations;
@@ -45,8 +46,7 @@ public class JiraVersionCreator
 
     var jiraProjectVersionRepairer = new JiraProjectVersionRepairer(_jiraProjectVersionService, _projectVersionFinder);
 
-    var versions = _projectVersionFinder.FindVersions(jiraProject, "(?s).*").ToList();
-    var jiraProjectVersion = versions.Where(x => x.name == versionNumber).DefaultIfEmpty().First();
+    var jiraProjectVersion = FindVersionWithVersionNumber(versionNumber);
 
     string createdVersionID;
 
@@ -69,5 +69,17 @@ public class JiraVersionCreator
     }
 
     return createdVersionID;
+  }
+
+  public JiraProjectVersion? FindVersionWithVersionNumber (string versionNumber)
+  {
+    var versions = _projectVersionFinder.FindVersions(_config.Jira.JiraProjectKey).ToList();
+     return versions.FirstOrDefault(x => x.name == versionNumber);
+  }
+  
+  public IReadOnlyList<JiraProjectVersion> FindAllVersionsStartingWithVersionNumber (string versionNumber)
+  {
+    var versions = _projectVersionFinder.FindVersions(_config.Jira.JiraProjectKey).ToList();
+    return versions.Where(x => x.name.StartsWith(versionNumber)).ToList();
   }
 }
--- a/ReleaseProcessAutomation/Jira/ServiceFacadeImplementations/JiraIssueService.cs
+++ b/ReleaseProcessAutomation/Jira/ServiceFacadeImplementations/JiraIssueService.cs
@@ -62,7 +62,7 @@ public class JiraIssueService
     }
   }
 
-  public IEnumerable<JiraToBeMovedIssue> FindAllNonClosedIssues (string versionId)
+  public IReadOnlyList<JiraToBeMovedIssue> FindAllNonClosedIssues (string versionId)
   {
     var jql = $"fixVersion={versionId} and resolution = \"unresolved\"";
     var resource = $"search?jql={jql}&fields=id,fixVersions,summary";
@@ -72,13 +72,28 @@ public class JiraIssueService
     return response.Data.Issues ?? throw new InvalidOperationException($"Could not get non closed issue data from jira with request '{resource}'");
   }
 
-  public IEnumerable<JiraToBeMovedIssue> FindAllClosedIssues (string versionId)
+  public IReadOnlyList<JiraToBeMovedIssue> FindAllClosedIssues (string versionId)
   {
     var jql = $"fixVersion={versionId} and resolution != \"unresolved\"";
-    var resource = $"search?jql={jql}&fields=id,fixVersions";
+    var resource = $"search?jql={jql}&fields=id,fixVersions,summary";
     var request = _jiraRestClientProvider.GetJiraRestClient().CreateRestRequest(resource, Method.GET);
 
     var response = _jiraRestClientProvider.GetJiraRestClient().DoRequest<JiraNonClosedIssues>(request, HttpStatusCode.OK);
     return response.Data.Issues ?? throw new InvalidOperationException($"Could not get closed issue data from jira with request '{resource}'");
+  }
+
+  public IReadOnlyList<JiraToBeMovedIssue> FindIssuesWithOnlyExactFixVersion (
+      IEnumerable<JiraProjectVersion> allVersions,
+      JiraProjectVersion exactFixVersion)
+  {
+    var allVersionLookup = allVersions.Select(v => v.id).ToHashSet();
+    allVersionLookup.Remove(exactFixVersion.id);
+
+    var allClosedIssues = FindAllClosedIssues(exactFixVersion.id);
+    var filteredIssues = allClosedIssues.Where(
+        i => i.Fields.FixVersions.TrueForAll(
+            v => !allVersionLookup.Contains(v.ID)));
+
+    return filteredIssues.ToList();
   }
 }
--- a/ReleaseProcessAutomation/Jira/ServiceFacadeImplementations/JiraProjectVersionFinder.cs
+++ b/ReleaseProcessAutomation/Jira/ServiceFacadeImplementations/JiraProjectVersionFinder.cs
@@ -20,6 +20,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Net;
 using System.Text.RegularExpressions;
+using JetBrains.Annotations;
 using ReleaseProcessAutomation.Jira.ServiceFacadeInterfaces;
 using ReleaseProcessAutomation.Jira.Utility;
 using RestSharp;
@@ -35,9 +36,13 @@ public class JiraProjectVersionFinder : IJiraProjectVersionFinder
     _jiraRestClientProvider = jiraRestClientProvider;
   }
 
-  public IEnumerable<JiraProjectVersion> FindVersions (string projectKey, string versionPattern)
+  public IEnumerable<JiraProjectVersion> FindVersions (string projectKey, string? versionPattern = null)
   {
     var versions = GetVersions(projectKey);
+
+    if (string.IsNullOrEmpty(versionPattern))
+      return versions;
+
     return versions.Where(
         v =>
         {
--- a/ReleaseProcessAutomation/Program.cs
+++ b/ReleaseProcessAutomation/Program.cs
@@ -18,6 +18,7 @@
 using System;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
+using ReleaseProcessAutomation.ReadInput;
 using Serilog;
 using Spectre.Console;
 
@@ -38,6 +39,11 @@ public static class Program
     try
     {
       app.Run(args);
+    }
+    catch (UserDoesNotWantToContinueException u)
+    {
+      Console.WriteLine(u.Message);
+      return -1;
     }
     catch (Exception e)
     {
--- a/ReleaseProcessAutomation/Steps/PipelineSteps/ReleaseAlphaBetaStep.cs
+++ b/ReleaseProcessAutomation/Steps/PipelineSteps/ReleaseAlphaBetaStep.cs
@@ -20,7 +20,6 @@ using System.Collections.Generic;
 using ReleaseProcessAutomation.Configuration.Data;
 using ReleaseProcessAutomation.Extensions;
 using ReleaseProcessAutomation.Git;
-using ReleaseProcessAutomation.Jira;
 using ReleaseProcessAutomation.ReadInput;
 using ReleaseProcessAutomation.Scripting;
 using ReleaseProcessAutomation.SemanticVersioning;
@@ -95,7 +94,7 @@ public class ReleaseAlphaBetaStep
 
     var nextJiraVersion = InputReader.ReadVersionChoice("Please choose next version (open JIRA issues get moved there):", nextPossibleJiraVersions);
 
-    _releaseVersionAndMoveIssuesSubStep.Execute(nextVersion, nextJiraVersion);
+    _releaseVersionAndMoveIssuesSubStep.Execute(nextVersion, nextJiraVersion, false, true);
 
     _msBuildCallAndCommit.CallMSBuildStepsAndCommit(MSBuildMode.PrepareNextVersion, nextVersion);
 
--- a/ReleaseProcessAutomation/Steps/SubSteps/ReleaseVersionAndMoveIssuesSubStep.cs
+++ b/ReleaseProcessAutomation/Steps/SubSteps/ReleaseVersionAndMoveIssuesSubStep.cs
@@ -1,5 +1,7 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Linq;
+using ReleaseProcessAutomation.Extensions;
 using ReleaseProcessAutomation.Jira;
 using ReleaseProcessAutomation.Jira.ServiceFacadeImplementations;
 using ReleaseProcessAutomation.ReadInput;
@@ -11,7 +13,7 @@ namespace ReleaseProcessAutomation.Steps.SubSteps;
 
 public interface IReleaseVersionAndMoveIssuesSubStep
 {
-  void Execute (SemanticVersion currentVersion, SemanticVersion nextVersion, bool squashUnreleased = false);
+  void Execute (SemanticVersion currentVersion, SemanticVersion nextVersion, bool squashUnreleased = false, bool movePreReleaseIssues = false);
 }
 
 public class ReleaseVersionAndMoveIssuesSubStep
@@ -38,7 +40,7 @@ public class ReleaseVersionAndMoveIssuesSubStep
     _jiraVersionReleaser = jiraVersionReleaser;
   }
 
-  public void Execute (SemanticVersion currentVersion, SemanticVersion nextVersion, bool squashUnreleased = false)
+  public void Execute (SemanticVersion currentVersion, SemanticVersion nextVersion, bool squashUnreleased = false, bool movePreReleaseIssues = false)
   {
     var currentVersionID = CreateVersion(currentVersion);
     var nextVersionID = CreateVersion(nextVersion);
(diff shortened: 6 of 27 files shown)


a-ctor (changes requested):

Frazerus:
Somehow the implementation didnt actually do what it was supposed to do, as the versions seem to not be in order? Don't know why the tests didn't fail in june, especially after looking at them again.

ReleaseProcessAutomation/Jira/JiraVersionCreator.cs
```diff
@@ -70,4 +70,16 @@ public string CreateNewVersionWithVersionNumber (string versionNumber)
 
     return createdVersionID;
   }
+
+  public JiraProjectVersion? FindVersionWithVersionNumber (string versionNumber)
+  {
+    var versions = _projectVersionFinder.FindVersions(_config.Jira.JiraProjectKey, "(?s).*").ToList();
```
a-ctor, line 76: General questions: Why do we need the `(?s)` in our regex?

ReleaseProcessAutomation/Jira/JiraVersionCreator.cs
```diff
@@ -70,4 +70,16 @@ public string CreateNewVersionWithVersionNumber (string versionNumber)
...
+     return versions.Where(x => x.name == versionNumber).DefaultIfEmpty().First();
```
a-ctor, line 77: We should write `.DefaultIfEmpty().First();` as `.FirstOrDefault()`

ReleaseProcessAutomation/Steps/PipelineSteps/ReleaseAlphaBetaStep.cs
```diff
@@ -94,7 +95,34 @@ public void Execute (SemanticVersion nextVersion, string? commitHash, bool pause
...
+        }
+        Console.WriteLine(
+            $"Do you want to move the closed versions with Jira fixVersion '{currentFullVersion}' to the newly released version '{nextVersion}'?");
+        if (InputReader.ReadConfirmation())
+          if (_jiraFunctionality.MoveClosedIssuesToNewVersion(currentFullVersion, nextVersion))
+          {
+            Console.WriteLine($"Could not move closed jira issues from version '{currentFullVersion}' to '{nextVersion}'. \nDo you wish to continue?");
+            if (!InputReader.ReadConfirmation())
+              return;
+          }
```
a-ctor, line 119: Style: If a statement spans across multiple lines braces should be added:

```suggestion
        if (InputReader.ReadConfirmation())
        {
          if (_jiraFunctionality.MoveClosedIssuesToNewVersion(currentFullVersion, nextVersion))
          {
            Console.WriteLine($"Could not move closed jira issues from version '{currentFullVersion}' to '{nextVersion}'. \nDo you wish to continue?");
            if (!InputReader.ReadConfirmation())
              return;
          }
        }
```

ReleaseProcessAutomation/Steps/PipelineSteps/ReleaseAlphaBetaStep.cs
```diff
@@ -94,7 +95,34 @@ public void Execute (SemanticVersion nextVersion, string? commitHash, bool pause
 
     var nextJiraVersion = InputReader.ReadVersionChoice("Please choose next version (open JIRA issues get moved there):", nextPossibleJiraVersions);
 
-    _ijIraFunctionality.CreateAndReleaseJiraVersion(nextVersion, nextJiraVersion);
+    _jiraFunctionality.CreateAndReleaseJiraVersion(nextVersion, nextJiraVersion);
+
+    var currentFullVersion = nextVersion.GetCurrentFullVersion();
```
a-ctor, line 100: Maybe we should move all of this logic into its own step. This would promote reuse, while also simplifying our logic inside as we can use `return` statements instead of all the nesting levels. What do you think?

ReleaseProcessAutomation/Jira/JiraFunctionality.cs
```diff
@@ -55,6 +61,39 @@ public void CreateAndReleaseJiraVersion (SemanticVersion currentVersion, Semanti
...
+
+  public IEnumerable<JiraToBeMovedIssue> GetClosedVersionsSummaryWithExactFixVersion (SemanticVersion version)
+  {
+    var currentVersions = _jiraVersionCreator.FindAllVersionsWithVersionNumber(version.ToString()).ToList() 
+                          ?? throw new InvalidOperationException($"Could not find version '{version}'.");
+
+    var fullVersion = _jiraVersionCreator.FindVersionWithVersionNumber(version.ToString()) 
+                      ?? throw new InvalidOperationException($"Could not find version '{version}'.");
+
+    currentVersions.Remove(currentVersions.Find(projectVersion => projectVersion.id.Equals(fullVersion.id))!);
```
a-ctor, line 87: We could simplify this statement using `RemoveAll(...)` as the id should be unique and only match one element.

ReleaseProcessAutomation/Jira/JiraFunctionality.cs
```diff
@@ -55,6 +61,39 @@ public void CreateAndReleaseJiraVersion (SemanticVersion currentVersion, Semanti
...
+    }
+    
+    var closedIssues = _jiraIssueService.FindAllClosedIssues(currentJiraVersion.id);
+    _jiraIssueService.MoveIssuesToVersion(closedIssues, currentJiraVersion.id, nextJiraVersion.id);
+    return true;
+  }
+
+  public IEnumerable<JiraToBeMovedIssue> GetClosedVersionsSummaryWithExactFixVersion (SemanticVersion version)
+  {
+    var currentVersions = _jiraVersionCreator.FindAllVersionsWithVersionNumber(version.ToString()).ToList()
```
a-ctor, line 81: We only use `currentVersions` to check if an element is in that list. So instead we could use a `HashSet<T>` which would simplify the removal of the `fullVersion` and also the checking logic. Since `JiraProjectVersion` does not implement equality we would need to use the id instead. Also there is a `.ToHashSet` method ;)

ReleaseProcessAutomation/Jira/JiraFunctionality.cs
```diff
@@ -55,6 +61,39 @@ public void CreateAndReleaseJiraVersion (SemanticVersion currentVersion, Semanti
...
+      _console.WriteLine($"Could not find current jira version '{currentVersion}' or next jira version '{nextVersion}'");
+      return false;
+    }
+    
+    var closedIssues = _jiraIssueService.FindAllClosedIssues(currentJiraVersion.id);
+    _jiraIssueService.MoveIssuesToVersion(closedIssues, currentJiraVersion.id, nextJiraVersion.id);
+    return true;
+  }
+
+  public IEnumerable<JiraToBeMovedIssue> GetClosedVersionsSummaryWithExactFixVersion (SemanticVersion version)
```
a-ctor, line 79: As it is, i don't understand what the method does just from the name. Maybe something like `GetIssuesForFixVersion` or something would be better.

ReleaseProcessAutomation/Steps/PipelineSteps/ReleaseAlphaBetaStep.cs
```diff
@@ -94,7 +95,34 @@ public void Execute (SemanticVersion nextVersion, string? commitHash, bool pause
...
+    try
+    {
+      var jiraClosedIssuesWithFixVersion = _jiraFunctionality.GetClosedVersionsSummaryWithExactFixVersion(currentFullVersion);
+      
+      if (jiraClosedIssuesWithFixVersion.Any())
+      {
+        foreach (var issue in jiraClosedIssuesWithFixVersion)
```
a-ctor, line 107: We are enumerating over the same enumerable twice, thus calculating the same thing twice. Maybe we should convert the enumerable to an array to prevent this.

ReleaseProcessAutomation/Jira/JiraFunctionality.cs
```diff
@@ -55,6 +61,39 @@ public void CreateAndReleaseJiraVersion (SemanticVersion currentVersion, Semanti
...
+    var nextJiraVersion = _jiraVersionCreator.FindVersionWithVersionNumber(nextVersion.ToString());
+    if(currentJiraVersion == null || nextJiraVersion == null)
+    {
+      _console.WriteLine($"Could not find current jira version '{currentVersion}' or next jira version '{nextVersion}'");
+      return false;
+    }
+    
+    var closedIssues = _jiraIssueService.FindAllClosedIssues(currentJiraVersion.id);
+    _jiraIssueService.MoveIssuesToVersion(closedIssues, currentJiraVersion.id, nextJiraVersion.id);
+    return true;
```
a-ctor, line 76: As it is used in `ReleaseAlphaBetaStep` this method should move all the issues listed in the `GetClosedVersionsSummaryWithExactFixVersion` method but the logic is a lot simpler. I think we are missing something here. Maybe it is better if we pass in the jira issues we want to have moved and the new version.

ReleaseProcessAutomation/Jira/JiraVersionCreator.cs
```diff
@@ -70,4 +70,16 @@ public string CreateNewVersionWithVersionNumber (string versionNumber)
...
+  }
+  
+  public IEnumerable<JiraProjectVersion> FindAllVersionsWithVersionNumber (string versionNumber)
+  {
+    var versions = _projectVersionFinder.FindVersions(_config.Jira.JiraProjectKey, "(?s).*").ToList();
+    return versions.Where(x => x.name.Contains(versionNumber));
```
a-ctor, line 83: Contains can have some nasty side-effects. I assume we want to get all the pre-release versions for a certain version. e.g. `3.1.0` -> `3.1.0-alpha.1`, `3.1.0-beta.2`, ... If this assumption is true I think we should use `.StartsWith` instead and change the name to reflect the change (e.g. `FindAllVersionsStartingWith`)

a-ctor (changes requested):

ReleaseProcessAutomation/Jira/JiraVersionCreator.cs
Frazerus: That is a great question. Tested it with and without it, no difference. 
The received data consists of individual objects already anyways. Maybe theres a possibility that the string is too long for jira and it cuts it off beforehand, but even with long names like 2.2.0-alpha.5 that didnt happen so... Possibly redundant, but it doesnt really hurt us and gives some safety against whatever happens in Jira.

ReleaseProcessAutomation/Steps/PipelineSteps/ReleaseAlphaBetaStep.cs
Frazerus: I agree that it should probably not be included inside the main part of the step, I would however not create a Jira Release Step just for this.

 I think a better option would be to just extract it into an extra method, either in the ReleaseAlphaBeta Step itself or the ReleaseProcessStepBase, as this would also promote reuse.
I dont think it would be beneficial to create a new step layer just for this one occurrence, as this would lead to like 5 new steps or some funky workaround just to get to the proper last step.

ReleaseProcessAutomation/Jira/JiraVersionCreator.cs
a-ctor: I guess that. `(?s)` is a non-capturing group with the `s` being an option for the group, but the group is empty. So it never matches anything and as a non-capturing group it won't create a group so there is no difference and we should remove it.

`Now the only thing left is `.*` which just matches everything. So we should remove the filter entirely to make it more readable and have an `.FindVersions(projectKey)` overload.

ReleaseProcessAutomation/Steps/PipelineSteps/ReleaseAlphaBetaStep.cs
a-ctor: I am fine with extracting this into a (base) method if we don't expect this functionality to be needed elsewhere. 

> as this would lead to like 5 new steps or some funky workaround just to get to the proper last step.

What do you mean by that? We would just move the code in the step and pass the step in the constructor. The benefit compared to the (base) method is that is that it is easier to reuse in another step. It is also easier to test as the logic can be tested in its isolated environment and the consuming steps can mock the inner step.

ReleaseProcessAutomation/Jira/JiraVersionCreator.cs
MichaelKetting: FYI: it was introduced here, but there's no reason for the single-line option:
https://github.com/re-motion/BuildTools/commit/a0424ddb530e9639c22baf86fd83ace21fa1ba1d#diff-e27d394fb9ff4d75b627934eec2ad5c69b6682139219f019f72209620ab4559bR54

ReleaseProcessAutomation/Steps/PipelineSteps/ReleaseAlphaBetaStep.cs
Frazerus: "Funky workaround" was maybe a bit wrong yeah. We would still need to create lots of extra things though:

As we do not want steps to go back on themselves, we would need to create a following step  just for the other things happening in all the "Release..." steps like the specific ms build settings as well as up to 3 possible following steps in one release part. 

This might be a thing for an extra ticket, as there is some redundance overall in the steps, there is some improvement potential there.

Frazerus: I will fix this by creating a base method for now, and will create a Task for a follow up ticket if that is ok for you.

Frazerus: As moving all of jira things just came up in another pr i just did it and moved everything in that branch ([RMSRCBUILD-43](https://github.com/re-motion/SharedSource-Build/pull/89))

Frazerus: We should probably finish 43 before we continue with this one as that should make this much more easy

Frazerus:
Viel Spaß beim Review, ich hatte keinen beim Rebasen >:(

ReleaseProcessAutomation/Jira/ServiceFacadeImplementations/JiraProjectVersionFinder.cs
```diff
@@ -48,6 +53,11 @@ public IEnumerable<JiraProjectVersion> FindVersions (string projectKey, string v
         });
   }
 
+  public IEnumerable<JiraProjectVersion> FindVersions (string projectKey)
+  {
+    return FindVersions(projectKey, null);
+  }
+
```
a-ctor, line 60: We can also remove this overload and use a default value instead. I am fine with both versions, benefit of the default values is that we don't have to introduce another interface method. As you like :)

ReleaseProcessAutomation/Jira/ServiceFacadeImplementations/JiraIssueService.cs
```diff
@@ -75,10 +75,24 @@ public IEnumerable<JiraToBeMovedIssue> FindAllNonClosedIssues (string versionId)
...
+  {
+    allVersions.RemoveWhere(projectVersion => projectVersion.id.Equals(exactFixVersion.id));
+
+    var allClosedIssues = FindAllClosedIssues(exactFixVersion.id);
+    var filteredIssues = allClosedIssues.Where(
+        i => i.Fields.FixVersions.TrueForAll(
+            v => !allVersions.Select(c => c.id).Contains(v.ID)));
+
+    return filteredIssues;
+  }
```
a-ctor, line 97: The way we use the `HashSet<JiraProjectVersion>` does not make any sense. We are using it like a list therefore making it worse than if we used just a `List<JiraProjectVersion>`. There are multiple solutions possible how would you solve the problem?

ReleaseProcessAutomation/Steps/SubSteps/ReleaseVersionAndMoveIssuesSubStep.cs
```diff
@@ -59,6 +61,60 @@ public void Execute (SemanticVersion currentVersion, SemanticVersion nextVersion
...
+    try
+    {
+      var currentVersions = _jiraVersionCreator.FindAllVersionsStartingWithVersionNumber(currentFullVersion)
+                            ?? throw new InvalidOperationException($"Could not find version '{currentFullVersion}'.");
+      ;
+
+      var allVersionsWithFullVersion = _jiraVersionCreator.FindVersionWithVersionNumber(currentFullVersion)
+                                       ?? throw new InvalidOperationException(
+                                           $"Could not find any version with version number '{currentFullVersion}'.");
+      ;
```
a-ctor, line 80: `;`

ReleaseProcessAutomation/Steps/SubSteps/ReleaseVersionAndMoveIssuesSubStep.cs
```diff
@@ -59,6 +61,60 @@ public void Execute (SemanticVersion currentVersion, SemanticVersion nextVersion
...
+    return !movePreReleaseIssues || MoveClosedIssuesFromPreviousVersion(currentVersion);
+  }
+
+  private bool MoveClosedIssuesFromPreviousVersion (SemanticVersion currentVersion)
+  {
+    var currentFullVersion = currentVersion.GetCurrentFullVersion().ToString();
+    try
+    {
+      var currentVersions = _jiraVersionCreator.FindAllVersionsStartingWithVersionNumber(currentFullVersion)
+                            ?? throw new InvalidOperationException($"Could not find version '{currentFullVersion}'.");
```
a-ctor, line 74: This will never return `null`. Either we remove the `throw` or change it to an empty check

ReleaseProcessAutomation/Steps/SubSteps/ReleaseVersionAndMoveIssuesSubStep.cs
```diff
@@ -38,7 +40,7 @@ public class ReleaseVersionAndMoveIssuesSubStep
     _jiraVersionReleaser = jiraVersionReleaser;
   }
 
-  public void Execute (SemanticVersion currentVersion, SemanticVersion nextVersion, bool squashUnreleased = false)
+  public bool Execute (SemanticVersion currentVersion, SemanticVersion nextVersion, bool squashUnreleased = false, bool movePreReleaseIssues = false)
```
a-ctor, line 43: I want to understand why we need to return a `bool` here and explore other options. Am I correct in assuming this was the related part to the "user abort exception question"? :D 

I dislike the boolean return as it is unclear what the value indicates. If we want to let the use abort the whole step then we should do it using a custom exception that we catch at the bottom and ignore/log.

ReleaseProcessAutomation/Steps/PipelineSteps/ReleaseAlphaBetaStep.cs
```diff
@@ -17,6 +17,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
```
a-ctor, line 20:
```suggestion
```

ReleaseProcessAutomation/Steps/PipelineSteps/ReleaseAlphaBetaStep.cs
```diff
@@ -95,13 +96,14 @@ public void Execute (SemanticVersion nextVersion, string? commitHash, bool pause
...
+    if (!_releaseVersionAndMoveIssuesSubStep.Execute(nextVersion, nextJiraVersion, squashUnreleased: false, movePreReleaseIssues: true))
+      return;
 
     _msBuildCallAndCommit.CallMSBuildStepsAndCommit(MSBuildMode.PrepareNextVersion, nextVersion);
 
     if (pauseForCommit)
       return;
 
-    _continueAlphaBetaStep.Execute(nextVersion, currentBranchName, currentBranchName, noPush);
+    _continueAlphaBetaStep.Execute(nextVersion, ancestor: currentBranchName, currentBranchName, noPush);
```
a-ctor, line 107: I would reset this change or do you think the annotation adds any value? If it does we should also include it at the other parameters since their intention is also unclear without an annotation IMHO

```suggestion
    _continueAlphaBetaStep.Execute(nextVersion, currentBranchName, currentBranchName, noPush);
```

a-ctor (changes requested):

ReleaseProcessAutomation/Jira/ServiceFacadeImplementations/JiraIssueService.cs
Frazerus: As im not that great with hashsets and new to LINQ i dont have a clue other than just removing the hashset and using a list altogether, but I dont think that that's the point ;)

ReleaseProcessAutomation/Steps/SubSteps/ReleaseVersionAndMoveIssuesSubStep.cs
Frazerus: Yeah, I didn't think it would get this much out of hand though.

ReleaseProcessAutomation/Program.cs
```diff
@@ -39,6 +40,11 @@ public static int Main (string[] args)
     {
       app.Run(args);
     }
+    catch (UserDoesNotWantToContinueException u)
+    {
+      Console.WriteException(u, ExceptionFormats.ShortenEverything);
+      return -1;
+    }
```
a-ctor, line 47: Do we want to print out a user abortion as exception? The message is always going to be empty. Maybe we just print the message as an info and supply a "User aborted the xxx operation" when throwing the exception

a-ctor (approved):

ReleaseProcessAutomation/Program.cs
Frazerus: Ah yeah I did think of that when writing that, but I had the plan of reusing the exception, but that would probably be better done in another ticket. Will create one.

Frazerus: Nevermind, I can't read. Thought of something else, will fix this and create a ticket for the other thing.

PR #93, opened 2022-06-30
Bug/RMSRCBUILD-196 release branch is not pushed to remote

Frazerus:
[Jira Issue 196](https://re-motion.atlassian.net/browse/RMSRCBUILD-196)

### DIFF
--- a/ReleaseProcessAutomation.UnitTests/Steps/Releases/ReleaseNonPreReleaseFromDevelopTests.cs
+++ b/ReleaseProcessAutomation.UnitTests/Steps/Releases/ReleaseNonPreReleaseFromDevelopTests.cs
@@ -41,6 +41,7 @@ internal class ReleaseNonPreReleaseFromDevelopTests
   private Mock<IMSBuildCallAndCommit> _msBuildInvokerMock;
   private Mock<IContinueReleaseOnMasterStep> _continueReleaseOnMasterMock;
   private Mock<IJiraFunctionality> _jiraFunctionalityMock;
+  private Mock<IPushNewReleaseBranchStep> _pushNewReleaseBranchMock;
 
   [SetUp]
   public void Setup ()
@@ -51,6 +52,7 @@ internal class ReleaseNonPreReleaseFromDevelopTests
     _continueReleaseOnMasterMock = new Mock<IContinueReleaseOnMasterStep>();
     _consoleMock = new Mock<IAnsiConsole>();
     _jiraFunctionalityMock = new Mock<IJiraFunctionality>();
+    _pushNewReleaseBranchMock = new Mock<IPushNewReleaseBranchStep>(MockBehavior.Strict);
   }
 
   [Test]
@@ -67,6 +69,7 @@ internal class ReleaseNonPreReleaseFromDevelopTests
         gitClientMock.Object,
         readInputMock.Object,
         _continueReleaseOnMasterMock.Object,
+        _pushNewReleaseBranchMock.Object,
         _config,
         _msBuildInvokerMock.Object,
         _consoleMock.Object,
@@ -98,6 +101,7 @@ internal class ReleaseNonPreReleaseFromDevelopTests
         gitClientStub.Object,
         readInputStub.Object,
         _continueReleaseOnMasterMock.Object,
+        _pushNewReleaseBranchMock.Object,
         _config,
         _msBuildInvokerMock.Object,
         _consoleMock.Object,
@@ -131,6 +135,7 @@ internal class ReleaseNonPreReleaseFromDevelopTests
         gitClientStub.Object,
         readInputStub.Object,
         _continueReleaseOnMasterMock.Object,
+        _pushNewReleaseBranchMock.Object,
         _config,
         _msBuildInvokerMock.Object,
         _consoleMock.Object,
@@ -143,7 +148,7 @@ internal class ReleaseNonPreReleaseFromDevelopTests
   }
 
   [Test]
-  public void Execute_NothingThrowsButStartReleasePhase_ShouldNotCallNextStep ()
+  public void Execute_NothingThrowsButStartReleasePhase_ShouldNotCallNextStepButPushReleaseBranchStep ()
   {
     var nextVersion = new SemanticVersion
                       {
@@ -157,6 +162,8 @@ internal class ReleaseNonPreReleaseFromDevelopTests
     gitClientStub.Setup(_ => _.IsOnBranch("develop")).Returns(true);
     gitClientStub.Setup(_ => _.DoesBranchExist($"release/v{nextVersion}")).Returns(false);
 
+    _pushNewReleaseBranchMock.Setup(_ => _.Execute($"release/v{nextVersion}", "develop"));
+    
     var readInputStub = new Mock<IInputReader>();
     readInputStub.Setup(_ => _.ReadVersionChoice(It.IsAny<string>(), It.IsAny<IReadOnlyCollection<SemanticVersion>>())).Returns(nextJiraVersion);
 
@@ -164,6 +171,7 @@ internal class ReleaseNonPreReleaseFromDevelopTests
         gitClientStub.Object,
         readInputStub.Object,
         _continueReleaseOnMasterMock.Object,
+        _pushNewReleaseBranchMock.Object,
         _config,
         _msBuildInvokerMock.Object,
         _consoleMock.Object,
@@ -172,7 +180,7 @@ internal class ReleaseNonPreReleaseFromDevelopTests
     step.Execute(nextVersion, "commitHash", true, false, false);
 
     _jiraFunctionalityMock.Verify(_ => _.CreateAndReleaseJiraVersion(nextVersion, nextJiraVersion, false), Times.Never);
-
+    _pushNewReleaseBranchMock.Verify(_ => _.Execute($"release/v{nextVersion}", "develop"));
     _continueReleaseOnMasterMock.Verify(_ => _.Execute(nextVersion, It.IsAny<bool>()), Times.Never);
   }
 
@@ -199,6 +207,7 @@ internal class ReleaseNonPreReleaseFromDevelopTests
         gitClientStub.Object,
         readInputStub.Object,
         _continueReleaseOnMasterMock.Object,
+        _pushNewReleaseBranchMock.Object,
         _config,
         _msBuildInvokerMock.Object,
         _consoleMock.Object,
--- a/ReleaseProcessAutomation.UnitTests/Steps/Releases/ReleasePatchStepTests.cs
+++ b/ReleaseProcessAutomation.UnitTests/Steps/Releases/ReleasePatchStepTests.cs
@@ -44,6 +44,7 @@ internal class ReleasePatchStepTests
   private Mock<IMSBuildCallAndCommit> _msBuildInvokerMock;
   private Mock<IContinueReleasePatchStep> _contineReleasePatchMock;
   private Mock<IJiraFunctionality> _jiraFunctionalityMock;
+  private Mock<IPushNewReleaseBranchStep> _pushNewReleaseBranchMock;
 
   [SetUp]
   public void Setup ()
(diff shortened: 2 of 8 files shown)


ReleaseProcessAutomation/Steps/PipelineSteps/PushNewReleaseBranch.cs
```diff
@@ -0,0 +1,31 @@
...
+using ReleaseProcessAutomation.Git;
+
+namespace ReleaseProcessAutomation.Steps.PipelineSteps;
+
+public interface IPushNewReleaseBranch
+{
+  void Execute (string releaseBranchName, string firstAncestorBranchName);
+}
+
+public class PushNewReleaseBranch
```
a-ctor, line 12: This class and its interface should have a `Step` suffix, otherwise they are located in the wrong folder

ReleaseProcessAutomation/Steps/PipelineSteps/PushNewReleaseBranch.cs
```diff
@@ -0,0 +1,31 @@
...
+    _gitClient = gitClient;
+    _config = config;
+  }
+  
+  public void Execute (string releaseBranchName, string firstAncestorBranchName)
+  {
+    var remoteNames = _config.RemoteRepositories.RemoteNames;
+    _gitClient.PushToRepos(remoteNames, releaseBranchName);
+    if(!string.IsNullOrEmpty(firstAncestorBranchName))
+      _gitClient.PushToRepos(remoteNames, firstAncestorBranchName);
```
a-ctor, line 29: It is kind of strange that we pass in two branch names and do the same thing (push) but the second one is optional. We should remove one of the parameters and let the caller decide how often they want to call us. The `ReleaseOnMasterStep` would call us twice while the `ReleasePatchStep` just once.

ReleaseProcessAutomation/Steps/PipelineSteps/ReleasePatchStep.cs
```diff
@@ -50,34 +51,37 @@ public class ReleasePatchStep : ReleaseProcessStepBase, IReleasePatchStep
...
+      if (!currentBranchName.StartsWith("master"))
       {
         const string message = "Cannot release a patch for master when not on master branch.";
         throw new InvalidOperationException(message);
       }
     }
     else
     {
-      if (!GitClient.IsOnBranch("hotfix/"))
+      if (!currentBranchName.StartsWith("hotfix/"))
```
a-ctor, line 84: Do we need those changes? They seem irrelevant to the issue at hand.

ReleaseProcessAutomation.Tests/Steps/Releases/ReleaseNonPreReleaseFromDevelopTests.cs
```diff
@@ -116,22 +120,22 @@ public void Execute_NothingThrows_ShouldCallNextStep()
...
-
-
+              gitClientStub.Object,
+              readInputStub.Object,
+              _continueReleaseOnMasterMock.Object,
+              _pushNewReleaseBranchMock.Object,
+              _config,
+              _msBuildInvokerMock.Object,
+              _consoleMock.Object);
+
```
a-ctor, line 130: Formatting was correct.

a-ctor (changes requested):
The branch name for this PR seems incorrect. There is also the branch `bugfix/RMSRCBUILD-196-release-branch-is-not-pushed-to-remote` which has the correct naming. We should use the branch with the correct name and delete the other branch. Be aware that this PR is closed when the branch is deleted so we should do this after all the reviews are done.

Also there are a few formatting issue left (e.g. `(_=>...` instead of `(_ => ...`). Please look through the changes and fix those formatting issues.

ReleaseProcessAutomation/Steps/PipelineSteps/PushNewReleaseBranch.cs
Frazerus: ReleasePatchStep would also need to call this twice, I think the only thing we could/should do is to remove the optionality of the second argument.

ReleaseProcessAutomation/Steps/PipelineSteps/ReleasePatchStep.cs
Frazerus: Well we need the current branch name already, so why not use it and save some git calls? I know this doesnt fit this issue, perfectly, it does come in contact with it however. 

Can still revert it if that is not enough of an argument ;)

a-ctor: `GitClient.IsOnBranch("bla")` and `currentBranch.StartsWith("bla")` do not seem like they test the same condition. We can keep the changes but we should address that issue.

ReleaseProcessAutomation/Steps/PipelineSteps/PushNewReleaseBranch.cs
a-ctor: If the second parameter is no longer optional then there is no reason why the method has to push two branches at the same time. In either case I think it is better to have the functionality of "pushing a branch" rather then "pushing a branch and optionally another one as well". I think this is better because the user can decide what they want to do. Otherwise, if you just want to push a single branch you either have to duplicate the logic or pass an empty string just for the sake of providing values.

ReleaseProcessAutomation/Steps/PipelineSteps/ReleasePatchStep.cs
Frazerus: Not quite sure how to address that issue, naming? Wrapper method? Git branches are defined by names in this implementation, so I think it should seem like startsWith checks the same, but IDK.

ReleaseProcessAutomation.UnitTests/Steps/Releases/ReleaseNonPreReleaseFromDevelopTests.cs
```diff
@@ -51,6 +52,7 @@ public void Setup ()
     _continueReleaseOnMasterMock = new Mock<IContinueReleaseOnMasterStep>();
     _consoleMock = new Mock<IAnsiConsole>();
     _jiraFunctionalityMock = new Mock<IJiraFunctionality>();
+    _pushNewReleaseBranchMock = new Mock<IPushNewReleaseBranchStep>();
```
a-ctor, line 55: I just realized that using a loose mock here is kind of bad. If the step is called when we don't want it to be called then such a test would not fail. We should use a strict mock and explicitly setup when we expect the call.

ReleaseProcessAutomation.UnitTests/Steps/Releases/ReleasePatchStepTests.cs
```diff
@@ -54,6 +55,7 @@ public void Setup ()
     _contineReleasePatchMock = new Mock<IContinueReleasePatchStep>();
     _consoleStub = new Mock<IAnsiConsole>();
     _jiraFunctionalityMock = new Mock<IJiraFunctionality>();
+    _pushNewReleaseBranchMock = new Mock<IPushNewReleaseBranchStep>();
```
a-ctor, line 58: See previous comment -> strict mock

ReleaseProcessAutomation/Steps/PipelineSteps/ReleasePatchStep.cs
```diff
@@ -52,36 +53,39 @@ public class ReleasePatchStep : ReleaseProcessStepBase, IReleasePatchStep
...
+    if (currentBranchName == null)
     {
       const string message = "Could not get the name of the current branch.";
       throw new InvalidOperationException(message);
     }
 
     if (onMaster)
     {
-      if (!GitClient.IsOnBranch("master"))
+      if (!currentBranchName.Equals("master"))
```
a-ctor, line 80: In C# the preferred way of checking equality is using `==`. Compared to Java when `.Equals` and `==` do different kinds of equality checks, `==` and `.Equals` are equivalent in C#. They both compare the value of the string and not the reference. `.Equals` is useful in situations where you want special string comparisons though, for example, case-insensitive string comparision, as you can supply a second argument for the type of comparison.

ReleaseProcessAutomation/Steps/PipelineSteps/ReleasePatchStep.cs
```diff
@@ -52,36 +53,39 @@ public class ReleasePatchStep : ReleaseProcessStepBase, IReleasePatchStep
...
       {
         const string message = "Cannot release a patch for master when not on master branch.";
         throw new InvalidOperationException(message);
       }
     }
     else
     {
-      if (!GitClient.IsOnBranch("hotfix/"))
+      if (!currentBranchName.StartsWith("hotfix/"))
```
a-ctor, line 88: This should also be changed to `==`

a-ctor (changes requested):

ReleaseProcessAutomation/Steps/PipelineSteps/ReleasePatchStep.cs
Frazerus: Ok, thank you for the input :)

ReleaseProcessAutomation/Steps/PipelineSteps/ReleasePatchStep.cs
Frazerus: Im not sure youre correct here, we dont know which version number will be appended to the hotfix/*** so we cant really check for equality

a-ctor: I should learn to read :D

ReleaseProcessAutomation.UnitTests/Steps/Releases/ReleasePatchStepTests.cs
```diff
@@ -118,29 +122,34 @@ public void Execute_OnHotfixWithoutErrors_CallsNextStep ()
...
     var nextVersion = new SemanticVersion { Patch = 1 };
     var nextJiraVersion = new SemanticVersion();
     _gitClientStub.Setup(_ => _.IsWorkingDirectoryClean()).Returns(true);
     _gitClientStub.Setup(_ => _.IsCommitHash("")).Returns(true);
     _gitClientStub.Setup(_ => _.IsOnBranch("hotfix/")).Returns(true);
-    _gitClientStub.Setup(_ => _.GetCurrentBranchName()).Returns("release/v0.0.1");
+    _gitClientStub.Setup(_ => _.GetCurrentBranchName()).Returns("hotfix/v0.0.1");
     var nextPossibleVersions = nextVersion.GetNextPossibleVersionsHotfix();
     _inputReaderStub.Setup(_ => _.ReadVersionChoice(It.IsAny<string>(), nextPossibleVersions)).Returns(nextJiraVersion);
+    _pushNewReleaseBranchMock.Setup(_=>_.Execute("release/v0.0.1", "hotfix/v0.0.1" ));
```
a-ctor, line 135:
```suggestion
    _pushNewReleaseBranchMock.Setup(_ => _.Execute("release/v0.0.1", "hotfix/v0.0.1" ));
```

ReleaseProcessAutomation.UnitTests/Steps/Releases/ReleasePatchStepTests.cs
```diff
@@ -118,29 +122,34 @@ public void Execute_OnHotfixWithoutErrors_CallsNextStep ()
...
         _contineReleasePatchMock.Object,
+        _pushNewReleaseBranchMock.Object,
         _consoleStub.Object,
         _jiraFunctionalityMock.Object);
 
     Assert.That(
         () => patchStep.Execute(nextVersion, "", true, false, false, false),
         Throws.Nothing);
+    
+    _pushNewReleaseBranchMock.Verify(_=>_.Execute("release/v0.0.1", "hotfix/v0.0.1" ));
```
a-ctor, line 152:
```suggestion
    _pushNewReleaseBranchMock.Verify(_ => _.Execute("release/v0.0.1", "hotfix/v0.0.1" ));
```

a-ctor (approved):
Two formatting remarks that would be nice if we fix them before merging. Also please squash everything into a single commit  `RMSRCBUILD-196 xxxx` before merging.

PR #95, opened 2022-08-11
Feature/RMSRCBUILD-292 allow multiple ms build path entries in the release process script.config

Frazerus:
https://re-motion.atlassian.net/browse/RMSRCBUILD-292

### DIFF
--- a/ReleaseProcessAutomation.UnitTests/Config/ConfigReaderTests.cs
+++ b/ReleaseProcessAutomation.UnitTests/Config/ConfigReaderTests.cs
@@ -17,6 +17,7 @@
 
 using System;
 using System.IO;
+using System.Linq;
 using NUnit.Framework;
 using ReleaseProcessAutomation.Configuration;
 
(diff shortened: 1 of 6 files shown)


ReleaseProcessAutomation/Configuration/ConfigReader.cs
```diff
@@ -59,9 +59,18 @@ public Config LoadConfig (string configPath)
...
+      if (!File.Exists(fullPath))
+        continue;
 
-    return config;
+      _log.Information("Using MSBuild from the path '{FullPath}'.", fullPath);
+      config.MSBuildSettings.ConfirmedMSBuildPath = fullPath;
+      return config;
+      
+    }
+    throw new ArgumentException("None of the configured ms build paths in the config exist.\nPlease configure a proper MSBuild path in the config.");
```
a-ctor, line 73:
```suggestion
    throw new ArgumentException("None of the configured MSBuild paths in the config exist.\nPlease configure a proper MSBuild path in the config.");
```

a-ctor (changes requested):

ReleaseProcessAutomation/Configuration/Data/Settings.cs
```diff
@@ -24,6 +24,8 @@ namespace ReleaseProcessAutomation.Configuration.Data;
 
 public class Settings
 {
+  public string ConfirmedMSBuildPath { get; set; }
```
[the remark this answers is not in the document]
Frazerus: Actually I think it would be better two have both versions! Have MsBuildPath as standard and add a secondary (like alternativeMsBuildPath) to the config, so we dont have unnecessary breaking changes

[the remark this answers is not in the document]
a-ctor: We can keep the `msBuildPath` naming for now to not break existing configs but I would still rename the property.

ReleaseProcessAutomation.UnitTests/Config/ConfigReaderTests.cs
```diff
@@ -75,4 +76,24 @@ public void GetConfigPathFromBuildProject_InCurrentWorkingDir_GetsProperName ()
...
+    var changedContent = configLines.Select(
+        l => l.Replace(
+            "<!--MsBuildPathTestPlaceholder-->",
+            "<msBuildPath>C:/DEFINETLYNOTANMSBUILDPATH/DEFINETLYNOTANMSBUILD.EXE</msBuildPath>")).ToArray();
+    File.WriteAllLines(configFilePath, changedContent);
+
+    var config = reader.LoadConfig(configFilePath);
+    
+    Assert.That(config.MSBuildSettings.MSBuildPath, Does.Contain("MSBuild.exe"));
+    Assert.That(File.Exists(config.MSBuildSettings.MSBuildPath), Is.True);
```
a-ctor, line 97: I am still not happy with the way we test this. We rely on a dummy value in the config file to exist and also rely on the fact that the XML is parsed so that the first element is taken. Also if the placeholder is changed or removed the test won't fail.

I propose hardcoding a minimal XML config file that we write to a file to pass the path to the reader. In the config file we have to placeholder values for the file paths and we fill them in before writing the file. We can use `Path.GetTempFile()` (or something like that) to create a file just for the test.

With that we can test the config reading in isolation without the dependency on the actual config.

ReleaseProcessAutomation.UnitTests/ReleaseProcessScript.Test.Config
```diff
@@ -24,7 +24,10 @@
         <fileName>Build/Customizations/Version.props</fileName>
     </developStableMergeIgnoreList>
     <msBuildSettings>
+        <!--MsBuildPathTestPlaceholder-->
         <msBuildPath>C:\Program Files\Microsoft Visual Studio\2022\Professional\MSBuild\Current\Bin\MSBuild.exe</msBuildPath>
+        <alternativeMSBuildPath>C:\Program Files\Microsoft Visual Studio\2022\Professional\MSBuild\Current\Bin\MSBuild.exe</alternativeMSBuildPath>
+        <alternativeMSBuildPath>C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Current\Bin\MSBuild.exe</alternativeMSBuildPath>
```
a-ctor, line 30: We do we have the same entry twice? Wouldn't it be enough to have one entry for the Professional edition?

ReleaseProcessAutomation.UnitTests/MsBuild/MsBuildInvokerTests.cs
```diff
@@ -135,7 +135,7 @@ public void InvokeMSBuildAndCommit_WrongMSBuildPathSpecified_EndsEarlyWithProper
     var act = msBuildInvoker.CallMSBuildStepsAndCommit(MSBuildMode.PrepareNextVersion, version);
 
     Assert.That(act, Is.EqualTo(-1));
-    Assert.That(testConsole.Output, Does.Contain($"The configured MSBuild path '{_config.MSBuildSettings.MSBuildPath}' does not exist.\nPlease configure a proper MSBuild path in the config.\nWill continue without invoking MSBuild.\n"));
+    Assert.That(testConsole.Output, Does.Contain($"The configured MSBuild path '{_config.MSBuildSettings.AlternativeMSBuildPath}' does not exist.\nPlease configure a proper MSBuild path in the config.\nWill continue without invoking MSBuild.\n"));
```
a-ctor, line 138: `AlternativeMSBuildPath` is an array so something is fishy here.

ReleaseProcessAutomation/Configuration/ConfigReader.cs
```diff
@@ -59,9 +59,26 @@ public Config LoadConfig (string configPath)
...
+      fullPath = Path.GetFullPath(alternativeMSBuildPath);
+      if (!File.Exists(fullPath))
+        continue;
+
+      _log.Information("Using MSBuild from the path '{FullPath}'.", fullPath);
+      config.MSBuildSettings.MSBuildPath = fullPath;
+      return config;
+      
+    }
+    throw new ArgumentException("None of the configured MSBuild paths in the config exist.\nPlease configure a proper MSBuild path in the config.");
```
a-ctor, line 81: We might be able to simplify this code into a single LINQ query that returns the first path that exists or null. I think this would improve the readability here.

a-ctor (changes requested):

ReleaseProcessAutomation/Configuration/Data/Settings.cs
```diff
@@ -23,7 +23,10 @@
 namespace ReleaseProcessAutomation.Configuration.Data;
 
 public class Settings
-{
+{
```
a-ctor, line 26:
```suggestion
{
```

ReleaseProcessAutomation/Configuration/Data/Settings.cs
```diff
@@ -23,7 +23,10 @@
 namespace ReleaseProcessAutomation.Configuration.Data;
 
 public class Settings
-{
+{  
   [XmlElement("msBuildPath")]
   public string MSBuildPath { get; set; }
+  
+  [XmlElement("alternativeMSBuildPath")]
+  public string[] AlternativeMSBuildPath { get; set; }
```
a-ctor, line 31: I just realized that we should use a plural here for both the XML name and the property name. Otherwise it is kind of strange to have multiple `<alternativeMSBuildPath>` next to each other

ReleaseProcessAutomation.UnitTests/Config/ConfigReaderTests.cs
```diff
@@ -75,4 +76,75 @@ public void GetConfigPathFromBuildProject_InCurrentWorkingDir_GetsProperName ()
 
     Assert.That(configPath, Is.EqualTo("Build/Customizations/ReleaseProcessScript.Test.config"));
   }
+
+  [Test]
```
a-ctor, line 80: Do we have a test that asserts the behavior that no MSBuild exe is found? If not we should add one

a-ctor (changes requested):

ReleaseProcessAutomation.UnitTests/Config/ConfigReaderTests.cs
Frazerus: Im questioning whether we need a test to figure out whether a proper MSBuild exe is found, or just whether something was added in the config that should point to some existent path. 

I already implemented the second one, but i wouldnt know whether or how to do the proper MSBuild.exe testing other than looking at whether the last part of the file path is called msbuild.exe

a-ctor (approved):

PR #103, opened 2023-01-13
RMSRCBUILD-302 Set ContinuousIntegrationBuild=true when building with build script

a-ctor:
https://re-motion.atlassian.net/browse/RMSRCBUILD-302

### DIFF
--- a/BuildScript.Nuke/Components/Tasks/CompileTask.cs
+++ b/BuildScript.Nuke/Components/Tasks/CompileTask.cs
@@ -145,6 +145,7 @@ internal static class CompileTask
           .SetProperty(MSBuildProperties.CompanyUrl, assemblyMetadata.CompanyUrl)
           .SetProperty(MSBuildProperties.ProductName, assemblyMetadata.ProductName)
           .SetProperty(MSBuildProperties.AssemblyOriginatorKeyFile, directories.SolutionKeyFile)
+          .SetProperty(MSBuildProperties.ContinuousIntegrationBuild, "true")
           .SetToolsVersion(project!.ToolsVersion)
           .SetConfiguration(configuration)
           .SetInformationalVersion(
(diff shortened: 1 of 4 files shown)


BuildScript.Nuke/Components/Tasks/CompileTask.cs
```diff
@@ -145,6 +145,7 @@ private static string GetToolPath (string msBuildNukePath, string msBuildPath, V
           .SetProperty(MSBuildProperties.CompanyUrl, assemblyMetadata.CompanyUrl)
           .SetProperty(MSBuildProperties.ProductName, assemblyMetadata.ProductName)
           .SetProperty(MSBuildProperties.AssemblyOriginatorKeyFile, directories.SolutionKeyFile)
+          .SetProperty(MSBuildProperties.ContinuousIntegrationBuild, "true")
```
MichaelKetting, line 148: Am not happy with the name. CI build is something else than a full build, from the semantics / intention. Technically, we have implemented them both via build script. Do you have additional naming suggestions?

BuildScript.Nuke/MSBuildProperties.cs
```diff
@@ -36,4 +36,5 @@ public static class MSBuildProperties
   public const string AssemblyInfoFile = "AssemblyInfoFile";
   public const string RestorePackagesConfig = "RestorePackagesConfig";
   public const string SolutionDir = "SolutionDir";
+  public const string ContinuousIntegrationBuild = "ContinuousIntegrationBuild";
```
MichaelKetting, line 39: should we switch those to MyConst = nameof(MyConst)?

MichaelKetting (changes requested):
Naming things

BuildScript.Nuke/Components/Tasks/CompileTask.cs
a-ctor: That is a dotnet SDK name :)

BuildScript.Nuke/MSBuildProperties.cs
a-ctor: I don't think that is a good idea since it makes it easier to break the loose binding with the MSBuild part by renaming the property.

MichaelKetting (approved):

PR #110, opened 2024-11-18
Feature/RMSRCBUILD 313 support SBOM generation

Frazerus:
https://re-motion.atlassian.net/browse/RMSRCBUILD-313


Build/Build.csproj
```diff
@@ -13,7 +13,8 @@
 
   <ItemGroup>
     <!-- Do not use a floating version here to prevent spirals of dooms when we introduce bugs. -->
-    <PackageReference Include="Remotion.BuildScript.Nuke" Version="4.0.0" />
+    <!-- <PackageReference Include="Remotion.BuildScript.Nuke" Version="4.0.0" /> TODO UNDO AFTERWARDS -->
```
MichaelKetting: comment to remember to TODO

Build/Build.csproj
```diff
@@ -14,6 +14,7 @@
   <ItemGroup>
     <!-- Do not use a floating version here to prevent spirals of dooms when we introduce bugs. -->
     <PackageReference Include="Remotion.BuildScript.Nuke" Version="4.0.0" />
+    <!-- <ProjectReference Include="../BuildScript.Nuke/BuildScript.Nuke.csproj" /> TODO UNDO AFTERWARDS -->
```
MichaelKetting: TODO reminder

BuildScript.Nuke/GenerateSbom/SolutionSbomGeneratorBuilder.cs
```diff
@@ -0,0 +1,74 @@
...
+  private AbsolutePath _pathToPackageJson;
+
+  public SolutionSbomGeneratorBuilder (Solution solution, AbsolutePath tempDirectory, AbsolutePath outputFile, string githubUsername, string githubAccessToken)
+  {
+    _solution = solution;
+    _tempDirectory = tempDirectory;
+    _outputFile = outputFile;
+
+    base.WithGithubCredentials(githubUsername, githubAccessToken);
+  }
```
MichaelKetting: I know, builder pattern and all that but is there actually a benefit to using WithGithubCredentials in the base class instead of keeping them in this builder?

And following that, is there a benefit in the base-class vs. interface?

MichaelKetting (commented):
Partial review. SolutionSbomGenerator is still to reivew.

BuildScript.Nuke/GenerateSbom/SolutionSbomGenerator.cs
```diff
@@ -0,0 +1,177 @@
...
+
+    Directory.CreateDirectory(_workingDirectory);
+
+    cycloneDxTool.Invoke(
+        arguments:
+        $"-o . --disable-package-restore --exclude-dev --exclude-test-projects -gu {_githubUsername} -gt {_githubAccessToken} -f \"{uncleanedSbomFilename}\" {_solution.Path}",
+        workingDirectory: _workingDirectory
+    );
+
+    Log.Information("Cleaning Sbom...");
```
MichaelKetting: we could write `SBOM` or `sbom` instead of `Sbom` in the various log messages. Pascal case isn't really common for this type of thing and SBOM isn't any different from JSON in the log message.

BuildScript.Nuke/GenerateSbom/SolutionSbomGenerator.cs
```diff
@@ -0,0 +1,177 @@
...
+        _projectBlacklist,
+        _workingDirectory / cleanedSbomFileName);
+
+    Log.Information("Created cleaned Sbom '{_outputFile}'.", _workingDirectory / cleanedSbomFileName);
+
+    if (_pathToPackageJson == null)
+    {
+      Log.Information("No package json file specified, will therefore not create combined Sbom.");
+      return;
+    }
```
MichaelKetting: I'm not sure we're setting `_outputFile` correctly when we are exiting here. I think it is only populated with the correct data when we do a dotnet and npm comine.

MichaelKetting: Code style suggestion: make two helper methods, one for .net and one for npm and do not do an early exit but a conditonal call to the npm stuff. that way, it is easier to see what happens/does not happen.

BuildScript.Nuke/GenerateSbom/SolutionSbomGenerator.cs
```diff
@@ -0,0 +1,177 @@
...
+    var httpRequestMessage = new HttpRequestMessage();
+    httpRequestMessage.Method = HttpMethod.Get;
+    httpRequestMessage.RequestUri = new Uri("https://api.github.com/repos/CycloneDX/cyclonedx-cli/releases");
+    httpRequestMessage.Headers.Add("User-Agent", "my-app");
+
+    var message = await httpClient.SendAsync(httpRequestMessage);
+    var result = await message.Content.ReadAsStringAsync();
+
+    var jsonData = (JArray)(JsonConvert.DeserializeObject(result)?? throw new InvalidOperationException($"Could not deserialize response from {httpRequestMessage.RequestUri}."));
+    var requiredAsset = jsonData.First?["assets"]?.First(jo => jo["name"]?.Value<string>() == "cyclonedx-win-x86.exe");
```
MichaelKetting: Should we pin the version of cyclonedx-cli? Then we can do https://github.com/CycloneDX/cyclonedx-cli/releases/download/v0.27.1/cyclonedx-win-x86.exe as a single call. Plus, we make sure we don't accidentally download bad stuff. A human would at least pause and think when there's weird stuff on the version. It's something I would feel better about given security concious design. Thoughts?

BuildScript.Nuke/Components/IGenerateSbom.cs
```diff
@@ -0,0 +1,56 @@
...
+using Nuke.Common.ProjectModel;
+using Nuke.Common.Tooling;
+using Remotion.BuildScript.GenerateSbom;
+using Serilog;
+
+namespace Remotion.BuildScript.Components;
+
+public interface IGenerateSbom: IBuild
+{
+  [NuGetPackage(
```
MichaelKetting, line 28: Should we pin the version of CycloneDX? would this pinning already happen via Directory.packages.props? or are we always pulling the latest version and hope for the best? Thoughts? see also comment on Cyclonedx-cli.

BuildScript.Nuke/GenerateSbom/SolutionSbomGenerator.cs
```diff
@@ -0,0 +1,177 @@
...
+
+    Log.Information("Merging npm and .net Sboms...");
+    PowerShellTasks.PowerShell(_ => _
+        .SetCommand($"{cycloneDXMergeToolPath} merge --input-files {cleanedSbomFileName} {npmSbomFileName} --output-file {_outputFile}")
+        .SetProcessWorkingDirectory(_workingDirectory));
+
+    Log.Information("Merged Sbom created at {0}", _outputFile);
+  }
+
+  private async Task DownloadCycloneDxMergeTool (AbsolutePath cycloneDXMergeToolPath)
```
MichaelKetting: Question:
CycloneDX.Utils appears to also offer Merge capabilities. WOuld this be a better choice than using the console application?
https://github.com/CycloneDX/cyclonedx-dotnet-library/blob/main/tests/CycloneDX.Utils.Tests/MergeTests.cs

MichaelKetting (changes requested):
I got a couple of questions. requesting changes because I'm not sure if I should just approve it. please check comments for what needs to be done and what can wait or should be discussed further as a future change.

BuildScript.Nuke/GenerateSbom/SolutionSbomGeneratorBuilder.cs
Frazerus: Yea, since we dont also have the csproj builder, we can just put them in this directly

BuildScript.Nuke/Components/IGenerateSbom.cs
Frazerus: Looking into Nuke's code base, this seems to check for the tool already being installed locally, so the nuget package seems to be required in a nuget package cache somewhere already. Not sure whether it also pulls the package if relevant, but it would probably just be good if we add the package manually somewhere, as on my pc I already have the tool. This would probably? fail on teamcity right now

Frazerus: Ok, looking deeper: It also looks into the project.assets.json file to see which version we use, so that should work

MichaelKetting: Okay, this means it uses the version we depend on in the csproj file of the Nuke-build-script and we need to add it there?

Frazerus: I dont know. Nuke does not say whether it uses the things from the solution we want to build or from the library we built around nuke.

Frazerus: Nevermind, I found it. we need to give the version of the tool in as a package reference or package download. So the things set in Buildscript.Nuke matter, the versions should be settable that way as well.

BuildScript.Nuke/BuildScript.Nuke.csproj
```diff
@@ -17,12 +17,18 @@
...
   </ItemGroup>
 
   <ItemGroup>
     <InternalsVisibleTo Include="BuildScript.Nuke.UnitTests" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="CycloneDX" Version="4.0.0" ExcludeAssets="all" />
+  </ItemGroup>
+
```
MichaelKetting, line 33:
```suggestion
    <PackageReference Include="CycloneDX" Version="4.0.0" ExcludeAssets="all" />
    <PackageReference Include="CycloneDX.Utils" Version="8.0.3" />
    <PackageReference Include="Nuke.Build" Version="8.0.0" />
    <PackageReference Include="Nuke.Common" Version="8.0.0" />
    <PackageReference Include="SbomCleaner.Library" Version="1.0.0" />
  </ItemGroup>

  <ItemGroup>
    <InternalsVisibleTo Include="BuildScript.Nuke.UnitTests" />
  </ItemGroup>
```
Would this be a good idea or was the split intentional because of the ExcludeAssets attribute?

BuildScript.Nuke/GenerateSbom/SolutionSbomGeneratorBuilder.cs
```diff
@@ -0,0 +1,78 @@
...
+
+public class SolutionSbomGeneratorBuilder : ISbomGeneratorBuilder
+{
+  private readonly Solution _solution;
+  private List<string> _packageBlackList = new();
+  private List<string> _projectBlackList = new();
+
+  private AbsolutePath _outputFile;
+  private AbsolutePath _tempDirectory;
+  private AbsolutePath _pathToPackageJson;
```
MichaelKetting:
```suggestion
  private readonly List<string> _packageBlackList = new();
  private readonly List<string> _projectBlackList = new();

  private readonly AbsolutePath _outputFile;
  private readonly AbsolutePath _tempDirectory;
  private AbsolutePath _pathToPackageJson;
```

Seems like a good idea, making those readonly while they're not mutable in the class? Thoughts?

BuildScript.Nuke/GenerateSbom/SolutionSbomGenerator.cs
```diff
@@ -0,0 +1,204 @@
...
+      File.Copy(_workingDirectory / cleanedSbomFileName, _outputFile);
+
+      Log.Information($"Cleaned sbom has been copied to '{_outputFile}'.");
+    }
+  }
+
+  private async Task MergeSboms (string cleanedSbomFileName, string npmSbomFileName)
+  {
+    var mainSbomFileStream = File.OpenRead(_workingDirectory / cleanedSbomFileName);
+    var npmSbomFileStream = File.OpenRead(_workingDirectory / npmSbomFileName);
```
MichaelKetting: should those have `using`?

BuildScript.Nuke/GenerateSbom/SolutionSbomGenerator.cs
```diff
@@ -0,0 +1,204 @@
...
+    {
+      Log.Information("No package json file specified, will therefore not create combined sbom.");
+
+      File.Copy(_workingDirectory / cleanedSbomFileName, _outputFile);
+
+      Log.Information($"Cleaned sbom has been copied to '{_outputFile}'.");
+    }
+  }
+
+  private async Task MergeSboms (string cleanedSbomFileName, string npmSbomFileName)
```
MichaelKetting: I don't see any `await` calls in this method. do we need `async Task`  instead of `void`?

BuildScript.Nuke/GenerateSbom/SolutionSbomGenerator.cs
```diff
@@ -0,0 +1,204 @@
...
+  {
+    var mainSbomFileStream = File.OpenRead(_workingDirectory / cleanedSbomFileName);
+    var npmSbomFileStream = File.OpenRead(_workingDirectory / npmSbomFileName);
+
+    var mainSbom = Serializer.Deserialize(mainSbomFileStream);
+    var npmsbom = Serializer.Deserialize(npmSbomFileStream);
+
+    npmSbomFileStream.DisposeAsync();
+    npmSbomFileStream.DisposeAsync();
```
MichaelKetting: do we need to do `DispostAsync` isntead of `Dispose`?

BuildScript.Nuke/GenerateSbom/SolutionSbomGenerator.cs
```diff
@@ -0,0 +1,204 @@
...
+
+
+    var combinedSbom =
+        CycloneDXUtils.HierarchicalMerge(new[] { mainSbom, npmsbom }, new Component
+                                                               {
+                                                                 Name = _solution.Name
+                                                               });
+
+    var outputFileStream = File.OpenWrite(_outputFile);
```
MichaelKetting: `using`?

BuildScript.Nuke/GenerateSbom/SolutionSbomGenerator.cs
```diff
@@ -0,0 +1,204 @@
...
+
+    PowerShellTasks.PowerShell(_ => _
+        .SetCommand($".\\node_modules\\.bin\\cyclonedx-npm.ps1 {packageJsonFolder /"package.json"} --omit dev --output-format xml --output-file '{npmSbomFileName}'")
+        .SetProcessWorkingDirectory(_workingDirectory));
+
+    Log.Information("Created npm sbom at '{npmSbomLocation}'.", _workingDirectory / npmSbomFileName);
+    return npmSbomFileName;
+  }
+
+  private async Task DownloadCycloneDxMergeTool (AbsolutePath cycloneDXMergeToolPath)
```
MichaelKetting: dead method, remove?

BuildScript.Nuke/GenerateSbom/SolutionSbomGenerator.cs
```diff
@@ -0,0 +1,204 @@
...
+    _workingDirectory = workingDirectory;
+    _outputFile = outputFile;
+    _projectBlacklist = projectBlacklist;
+    _packageBlackList = packageBlackList;
+    _pathToPackageJson = pathToPackageJson;
+    _githubUsername = githubUsername;
+    _githubAccessToken = githubAccessToken;
+  }
+
+  public async Task Generate (Tool cycloneDxTool)
```
MichaelKetting: should this method remain `async Task` (i.e. required by Nuke) after the MergeSboms is no longer async?

BuildScript.Nuke/GenerateSbom/SolutionSbomGenerator.cs
```diff
@@ -0,0 +1,204 @@
...
+        _projectBlacklist,
+        _workingDirectory / cleanedSbomFileName);
+
+    Log.Information("Created cleaned sbom '{_outputFile}'.", _workingDirectory / cleanedSbomFileName);
+
+    if (_pathToPackageJson != null)
+    {
+      var npmSbomFileName = CreateNPMSbom();
+
+      await MergeSboms(cleanedSbomFileName, npmSbomFileName);
```
MichaelKetting: design recoommendation: even though we have output-file as a memberfield, it is strange to not pass it here. this is becasue we would get a symetic api otherwise, always telling the reader, where the output file goes.

MichaelKetting (changes requested):

BuildScript.Nuke/BuildScript.Nuke.csproj
Frazerus: I think it makes sense to split the normal dependencies and the one for the Tool which is not used otherwise.

MichaelKetting: okay, great. please add a comment inside the second itemgroup so we know why it's done.

MichaelKetting (approved):

PR #111, opened 2024-12-16
Feature/RMSRCBUILD 315 add solution version to generated SBOM

Frazerus:
https://re-motion.atlassian.net/browse/RMSRCBUILD-315

### DIFF
--- a/BuildScript.Nuke/GenerateSbom/SolutionSbomGenerator.cs
+++ b/BuildScript.Nuke/GenerateSbom/SolutionSbomGenerator.cs
@@ -116,7 +116,7 @@ public class SolutionSbomGenerator: ISbomGenerator
     {
       Log.Information("No package json file specified, will therefore not create combined sbom.");
 
-      Directory.CreateDirectory(_outputFile.Parent);
+      _outputFile.Parent.CreateDirectory();
 
       File.Copy(_workingDirectory / cleanedSbomFileName, _outputFile, overwrite: true);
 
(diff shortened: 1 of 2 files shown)


BuildScript.Nuke/GenerateSbom/SolutionSbomGenerator.cs
```diff
@@ -109,6 +114,9 @@ public void Generate (Tool cycloneDxTool)
     {
       Log.Information("No package json file specified, will therefore not create combined sbom.");
 
+      Directory.CreateDirectory(_outputFile.Parent);
```
a-ctor: NUKE has `AbsolutePath.CreateDirectory()` so we could write `_outputFile.Parent.CreateDirectory()` instead.

BuildScript.Nuke/BuildScript.Nuke.csproj
```diff
@@ -31,4 +31,17 @@
...
+      <PackagePath>contentFiles\any\any\BuildHost-net472;content/BuildHost-net472</PackagePath>
+      <PackageCopyToOutput>true</PackageCopyToOutput>
+    </Content>
+    <Content Include="bin/Debug/BuildHost-netcore/**/*.*">
+      <Pack>true</Pack>
+      <PackagePath>contentFiles\any\any\BuildHost-netcore;content/BuildHost-netcore</PackagePath>
+      <PackageCopyToOutput>true</PackageCopyToOutput>
+    </Content>
+  </ItemGroup>
+
```
a-ctor: I think we are getting rid of that in anther PR so we just need to make sure it does not get reintroduced afterwards here. Maybe we should remove it here so that we can merge this PR?

a-ctor (approved):

Frazerus:
Only merge after #113

a-ctor (approved):

PR #112, opened 2024-12-17
RMSRCBUILD-309 Add support for DependDB

a-ctor:
https://re-motion.atlassian.net/browse/RMSRCBUILD-309

### DIFF
--- a/BuildScript.Nuke/Components/IDependDB.cs
+++ b/BuildScript.Nuke/Components/IDependDB.cs
@@ -1,0 +1,176 @@
+﻿// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+//
+// See the NOTICE file distributed with this work for additional information
+// regarding copyright ownership.  rubicon licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may not use this
+// file except in compliance with the License.  You may obtain a copy of the
+// License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+// License for the specific language governing permissions and limitations
+// under the License.
+
+using System;
+using System.IO;
+using System.Linq;
+using JetBrains.Annotations;
+using Nuke.Common;
+using Nuke.Common.IO;
+using Nuke.Common.Tooling;
+using Serilog;
+
+namespace Remotion.BuildScript.Components;
+
+public interface IDependDB : IBaseBuild, IBuildMetadata, IPack
+{
+  [NuGetPackage(
+      packageId: "DependDB.BuildProcessor.NuGetPreProcessor",
+      packageExecutable: "DependDB.BuildProcessor.NuGetPreProcessor.exe")]
+  Tool NugetPreProcessor => TryGetValue(() => NugetPreProcessor);
+
+  [NuGetPackage(
+      packageId: "DependDB.BuildProcessor",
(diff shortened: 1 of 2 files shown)


BuildScript.Nuke/Components/IDependDB.cs
```diff
@@ -0,0 +1,173 @@
...
+              if (DependDBNugetConfig != null)
+                  arguments = arguments.Add("--nuget-config={value}", MakeAbsolutePath(DependDBNugetConfig));
+
+              foreach (var trackedReference in trackedReferences)
+                  arguments = arguments.Add("--analzyer-tracked-reference={value}", trackedReference);
+
+              foreach (var nugetSource in nugetSources)
+                  arguments = arguments.Add("--additional-nuget-source={value}", nugetSource);
+
+              var argumentStringHandler = new ArgumentStringHandler(1, 0, out var _);
```
Frazerus, line 150: Please add a description for what in the heck is going on here

BuildScript.Nuke/Components/IDependDB.cs
```diff
@@ -0,0 +1,173 @@
...
+              NugetPreProcessor(
+                      argumentStringHandler,
+                      workingDirectory: Solution.Directory);
+
+              BuildProcessor(new Arguments().Add(buildProcessorFile).RenderForExecution());
+
+
+              string MakeAbsolutePath (string value)
+              {
+                  if (value.StartsWith("~/"))
```
Frazerus, line 162: Is this standard around this solution to have non absolute paths starting with ~? Or is it just here?

BuildScript.Nuke/Components/IDependDB.cs
```diff
@@ -0,0 +1,173 @@
...
+              Log.Information($"  OwnPackagesDirectory: '{ownPackagesDirectory}'");
+              Log.Information($"  {nameof(DependDBProjectName)}: '{DependDBProjectName}'");
+              Log.Information($"  {nameof(DependDBProjectBranch)}: '{DependDBProjectBranch}'");
+              Log.Information($"  {nameof(DependDBProjectVersion)}: '{projectVersion}'");
+              Log.Information($"  {nameof(DependDBRetentionTime)}: '{retentionTime}'");
+              Log.Information($"  {nameof(DependDBProjectImportNotificationMailAddress)}: '{DependDBProjectImportNotificationMailAddress}'");
+              Log.Information($"  {nameof(DependDBProjectImportNotificationMailAddressCc)}: '{DependDBProjectImportNotificationMailAddressCc}'");
+              Log.Information($"  {nameof(VcsUrlTemplate)}: '{VcsUrlTemplate}'");
+              Log.Information($"  {nameof(DependDBTargetFramework)}: '{DependDBTargetFramework}'");
+              Log.Information($"  {nameof(DependDBTrackedReferences)}: [{string.Join(", ", trackedReferences.Select(e => $"'{e}'"))}]");
```
Frazerus, line 118: Possibly easier to read with \n instead of ',' to split it?

Frazerus (approved):

BuildScript.Nuke/Components/IDependDB.cs
a-ctor: The thing is that we can't have absolute paths in the configuration files that are located in the repo because the checkout directory might change. Relative paths might be interpreted incorrectly using `Path.CurrentDirectory` so I decided to use the ASP.NET pattern `~/` to indicate the application root (-> the root of the repo).

BuildScript.Nuke/Components/IDependDB.cs
a-ctor: The entries should not be too long. It would improve the output a bit but the code here gets a bit more unreadable because of the required indent. The output here is just meant to aid in debugging so I think the current solution is fine.

PR #116, opened 2025-01-15
RMSRCBUILD-320 Support sbom generation with json

Frazerus:
https://re-motion.atlassian.net/browse/RMSRCBUILD-320

> [!NOTE]
> Manual review of output files

### DIFF
--- a/BuildScript.Nuke/GenerateSbom/SolutionSbomGenerator.cs
+++ b/BuildScript.Nuke/GenerateSbom/SolutionSbomGenerator.cs
@@ -17,8 +17,8 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using CycloneDX.Models;
 using CycloneDX.Utils;
-using CycloneDX.Xml;
 using Nuke.Common.IO;
 using Nuke.Common.ProjectModel;
 using Nuke.Common.Tooling;
@@ -26,6 +26,8 @@ using Nuke.Common.Tools.Npm;
 using Nuke.Common.Tools.PowerShell;
 using SbomCleaner.Library;
 using Serilog;
+using XMLSerializer = CycloneDX.Xml.Serializer;
+using JsonSerializer = CycloneDX.Json.Serializer;
 using Tool = Nuke.Common.Tooling.Tool;
 
 namespace Remotion.BuildScript.GenerateSbom;
@@ -92,7 +94,7 @@ public class SolutionSbomGenerator: ISbomGenerator
 
     Log.Information("Cleaning sbom...");
 
-    var cleanedSbomFileName = "cleaned-sbom.xml";
+    var cleanedSbomFile = _workingDirectory / "cleaned-sbom.xml";
     var cleaner = new Cleaner();
 
     cleaner.CleanSolutionSbom(
@@ -100,15 +102,15 @@ public class SolutionSbomGenerator: ISbomGenerator
         projects,
         _packageBlackList,
         _projectBlacklist,
-        _workingDirectory / cleanedSbomFileName);
+        cleanedSbomFile);
 
-    Log.Information("Created cleaned sbom '{_outputFile}'.", _workingDirectory / cleanedSbomFileName);
+    Log.Information("Created cleaned sbom '{_outputFile}'.", cleanedSbomFile);
 
     if (_pathToPackageJson != null)
     {
       var npmSbomFileName = CreateNPMSbom();
 
-      MergeSboms(cleanedSbomFileName, npmSbomFileName, _outputFile);
+      MergeSboms(cleanedSbomFile, npmSbomFileName, _outputFile);
     }
     else
     {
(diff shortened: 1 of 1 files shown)


BuildScript.Nuke/GenerateSbom/SolutionSbomGenerator.cs
```diff
@@ -116,25 +118,56 @@ public void Generate (Tool cycloneDxTool, IReadOnlyCollection<ProjectInfo> proje
 
       _outputFile.Parent.CreateDirectory();
 
-      File.Copy(_workingDirectory / cleanedSbomFileName, _outputFile, overwrite: true);
+      //The sbom cleaner only works with xml files, so we need to convert the sbom if we want a different format.
+      var outputSbomFileToCopy = _workingDirectory / cleanedSbomFileName;
+      if (_outputFile.Extension.EndsWith("json"))
```
a-ctor: `.EndsWith` feels a bit strange here. Why do we want to support extensions like `blablajson`? I would suggest using `_outputFile.Extension == ".json"` instead

BuildScript.Nuke/GenerateSbom/SolutionSbomGenerator.cs
```diff
@@ -116,25 +118,56 @@ public void Generate (Tool cycloneDxTool, IReadOnlyCollection<ProjectInfo> proje
...
+    var finalBom = XMLSerializer.Deserialize(cleanedBomStream);
+
+    var outputSbomJson = _workingDirectory / "output-sbom.json";
+    using var outputSbomJsonStream = File.Create(outputSbomJson);
+
+    // ReSharper disable once AccessToDisposedClosure -> We run synchronously, the dispose therefore always comes after the stream has been written to.
+    Task.Run(async () => await JsonSerializer.SerializeAsync(finalBom, outputSbomJsonStream)).GetAwaiter().GetResult();
+
+    outputSbomFileToCopy = outputSbomJson;
+    return outputSbomFileToCopy;
```
a-ctor:
```suggestion
    return outputSbomJson;
```

BuildScript.Nuke/GenerateSbom/SolutionSbomGenerator.cs
```diff
@@ -116,25 +118,56 @@ public void Generate (Tool cycloneDxTool, IReadOnlyCollection<ProjectInfo> proje
...
+  private AbsolutePath DuplicateAsJsonSbom (AbsolutePath outputSbomFileToCopy)
+  {
+    using var cleanedBomStream = File.OpenRead(outputSbomFileToCopy);
+    var finalBom = XMLSerializer.Deserialize(cleanedBomStream);
+
+    var outputSbomJson = _workingDirectory / "output-sbom.json";
+    using var outputSbomJsonStream = File.Create(outputSbomJson);
+
+    // ReSharper disable once AccessToDisposedClosure -> We run synchronously, the dispose therefore always comes after the stream has been written to.
+    Task.Run(async () => await JsonSerializer.SerializeAsync(finalBom, outputSbomJsonStream)).GetAwaiter().GetResult();
```
a-ctor: We don't need the async lambda + `Task.Run` to wait synchronously. This way, we should also no longer need "pragma"

```suggestion
    JsonSerializer.SerializeAsync(finalBom, outputSbomJsonStream).GetAwaiter().GetResult();
```

BuildScript.Nuke/GenerateSbom/SolutionSbomGenerator.cs
```diff
@@ -116,25 +118,56 @@ public void Generate (Tool cycloneDxTool, IReadOnlyCollection<ProjectInfo> proje
...
       Log.Information($"Cleaned sbom has been copied to '{_outputFile}'.");
     }
   }
 
+  private AbsolutePath DuplicateAsJsonSbom (AbsolutePath outputSbomFileToCopy)
+  {
+    using var cleanedBomStream = File.OpenRead(outputSbomFileToCopy);
+    var finalBom = XMLSerializer.Deserialize(cleanedBomStream);
+
+    var outputSbomJson = _workingDirectory / "output-sbom.json";
```
a-ctor: It feels strange to me that we get the input file via parameter but the output file is calculated from field state. I would suggest introducing a parameter for `outputSbomJson` and then we should be able to make this method `static` as well.

BuildScript.Nuke/GenerateSbom/SolutionSbomGenerator.cs
```diff
@@ -116,25 +118,56 @@ public void Generate (Tool cycloneDxTool, IReadOnlyCollection<ProjectInfo> proje
...
-    Serializer.Serialize(combinedSbom, outputFileStream);
+    if (outputFile.Extension.EndsWith("xml"))
+    {
+      XMLSerializer.Serialize(combinedSbom, outputFileStream);
+    }
+    if (outputFile.Extension.EndsWith("json"))
+    {
+      // ReSharper disable once AccessToDisposedClosure -> We run synchronously, the dispose therefore always comes after the stream has been written to.
+      Task.Run(async () => await JsonSerializer.SerializeAsync(combinedSbom, outputFileStream)).GetAwaiter().GetResult();
+    }
```
a-ctor: See other comments about file extension and async awaiting

a-ctor (changes requested):

BuildScript.Nuke/GenerateSbom/SolutionSbomGenerator.cs
```diff
@@ -116,25 +118,55 @@ public void Generate (Tool cycloneDxTool, IReadOnlyCollection<ProjectInfo> proje
 
       _outputFile.Parent.CreateDirectory();
 
-      File.Copy(_workingDirectory / cleanedSbomFileName, _outputFile, overwrite: true);
+      //The sbom cleaner only works with xml files, so we need to convert the sbom if we want a different format.
+      //Given a requested xml file, we can just copy
+      var outputSbomFileToCopy = _workingDirectory / cleanedSbomFileName;
```
a-ctor: We use `_workingDirectory / cleanedSbomFileName` two times to invoke the cleaner. As those expressions must match this one, I would move this variable declaration up above the Cleaner invocation and replace the other usages

BuildScript.Nuke/GenerateSbom/SolutionSbomGenerator.cs
```diff
@@ -116,25 +118,55 @@ public void Generate (Tool cycloneDxTool, IReadOnlyCollection<ProjectInfo> proje
...
+    var combinedSbom = CycloneDXUtils.FlatMerge([mainSbom, npmSbom], component);
 
     using var outputFileStream = File.Create(outputFile);
 
-    Serializer.Serialize(combinedSbom, outputFileStream);
+    if (outputFile.Extension.EndsWith("xml"))
+    {
+      XMLSerializer.Serialize(combinedSbom, outputFileStream);
+    }
+    if (outputFile.Extension.EndsWith("json"))
```
a-ctor: As mentioned before I would prefer `outputFile.Extension == ".ext"` to remove the need to think about why we are using `.EndsWith(...)`

a-ctor (approved):

BuildScript.Nuke/GenerateSbom/SolutionSbomGenerator.cs
Frazerus: Ah I only inspected it in the other case. thanks for finding this

Frazerus:
Too many changes for a simple go ahead, therefore a rerequest @a-ctor

BuildScript.Nuke/GenerateSbom/SolutionSbomGenerator.cs
```diff
@@ -92,55 +94,84 @@ public void Generate (Tool cycloneDxTool, IReadOnlyCollection<ProjectInfo> proje
...
+    using var outputSbomJsonStream = File.Create(finalJsonSbomLocation);
 
-    var combinedSbom = CycloneDXUtils.HierarchicalMerge([mainSbom, npmSbom], mainSbom.Metadata.Component);
+    JsonSerializer.SerializeAsync(finalBom, outputSbomJsonStream).GetAwaiter().GetResult();
+  }
+
+  private void MergeSboms (AbsolutePath cleanedSbomFile, AbsolutePath npmSbomFile, AbsolutePath outputFile)
+  {
+    using var mainSbomFileStream = File.OpenRead(cleanedSbomFile);
+    using var npmSbomFileStream = File.OpenRead( npmSbomFile);
```
a-ctor:
```suggestion
    using var npmSbomFileStream = File.OpenRead(npmSbomFile);
```

BuildScript.Nuke/GenerateSbom/SolutionSbomGenerator.cs
```diff
@@ -92,55 +94,84 @@ public void Generate (Tool cycloneDxTool, IReadOnlyCollection<ProjectInfo> proje
...
+
+    var mainSbom = XMLSerializer.Deserialize(mainSbomFileStream);
+    var npmSbom = XMLSerializer.Deserialize(npmSbomFileStream);
+
+    var component = new Component { Type = Component.Classification.Library, Name = _solution.Name + "-root-component", Version = _version };
```
a-ctor:
```suggestion
    var component = new Component
    {
      Type = Component.Classification.Library,
      Name = _solution.Name + "-root-component",
      Version = _version
    };
```

a-ctor (approved):

